PR #11, opened 2024-07-04
feat: create map wrapper for OpenLayers

mlu-ssoft:


libs/map/ui-map/.eslintrc.json
```diff
@@ -0,0 +1,18 @@
...
+      "files": ["*.ts", "*.tsx", "*.js", "*.jsx"],
+      "rules": {}
+    },
+    {
+      "files": ["*.ts", "*.tsx"],
+      "rules": {}
+    },
+    {
+      "files": ["*.js", "*.jsx"],
+      "rules": {}
```
KatarzynaPaszko, line 15: Is it needed? Will we have separate linting rules for libs? I would prefer to keep same rules for entire project

KatarzynaPaszko (commented):
We need to agree on testing steps for each task. Maybe we could add testing steps to PR description on how do you want me to verify the task is correct? 

I have run branch locally and I see map, run tests and linting? Do I need to check something more?

PR #14, opened 2024-07-15
UKRIW-141 - Create design tokens

KatarzynaPaszko:


apps/eodh-fe/src/styles/main.css
```diff
@@ -1,3 +1,7 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+body {
```
mlu-ssoft, line 5: shouldn't be `*`? double check if font will be applied to every element

libs/shared/theme/src/index.ts
```diff
@@ -0,0 +1 @@
+export * from './lib/tokens';
```
mlu-ssoft, line 1: do not export everything, export only things that are needed outside each library.
There are internal (inside library) and external (outside library) imports.
Not everything should be accessible outside library

libs/shared/theme/src/lib/tokens.js
```diff
@@ -0,0 +1,52 @@
...
+      function: '#FF0000',
+      area: '#3CAF5C',
+      data: '#FF7A00',
+      date: '#00B2FF',
+    },
+  },
+  // eg. text-small-semibold
+  fontSize: {
+    'small-semibold': ['8px', { lineHeight: '12px', fontWeight: '400' }],
+    'small-medium': ['9px', { lineHeight: '13.5px', fontWeight: '500' }],
```
mlu-ssoft, line 43: does poppins supports `400/500/600/700` font weight?

mlu-ssoft (changes requested):

apps/eodh-fe/src/styles/main.css
KatarzynaPaszko: It works fine with `body` due to inheritance. 
For [various reasons](https://stackoverflow.com/questions/10041706/applying-a-single-font-to-an-entire-website-with-css) it is better to apply things to `body` instead of `*` selector

libs/shared/theme/src/lib/tokens.js
KatarzynaPaszko: I double checked, and it supports just 2 font weights, so I will update this file accordingly

KatarzynaPaszko: I ended up adding installing Poppins font in repo, as Toby requested all fin-weight support. So it will support now `400/500/600/700`

apps/eodh-fe/vite.config.ts
```diff
@@ -17,6 +18,12 @@ export default defineConfig({
   server: {
     port: 4200,
     host: 'localhost',
+    fs: {
+      allow: [
+        // Dynamically resolve the project root
+        join(__dirname, '..', '..'),
```
mlu-ssoft, line 24: what this is doing?

apps/storybook-host/.storybook/preview.ts
```diff
@@ -1 +1,5 @@
 import './main.css';
+import '@fontsource/poppins/400.css';
```
mlu-ssoft, line 2: move those into `main.css` file

libs/shared/theme/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export { fontSize } from './lib/font-size';
+export { colors } from './lib/colors';
```
mlu-ssoft, line 2: maybe just export `themeConfig` which will have whole configuration for theme (without splitting it into multiple variables)

libs/shared/theme/src/lib/colors.js
```diff
@@ -0,0 +1,127 @@
+const colors = {
+  lightMode: {
+    // eg. bg-primary-light
```
mlu-ssoft, line 3: remove

libs/shared/theme/src/lib/colors.js
```diff
@@ -0,0 +1,127 @@
+const colors = {
+  lightMode: {
```
mlu-ssoft, line 2: shouldn't it be just `light`?

libs/shared/theme/src/lib/font-size.js
```diff
@@ -0,0 +1,12 @@
+const fontSize = {
+  // eg. text-small-semibold
```
mlu-ssoft, line 2: remove

package.json
```diff
@@ -26,6 +26,7 @@
   "private": true,
   "dependencies": {
     "@date-io/date-fns": "3.0.0",
+    "@fontsource/poppins": "^5.0.14",
```
mlu-ssoft, line 29: remove `^` and run npm install once again

mlu-ssoft (changes requested):

apps/eodh-fe/vite.config.ts
KatarzynaPaszko: It was added once running some of Nx commands. 

As I [read in docs](https://vitejs.dev/config/server-options#server-fs-allow), the purpose of this `fs.allow `configuration is to enable Vite to access files outside of the immediate project directory. In an Nx monorepo, this is particularly useful because it allows Vite to resolve dependencies and shared code that might exist in other parts of your workspace

libs/shared/theme/src/lib/colors.js
KatarzynaPaszko: I wanted it to be self explanatory, but iIf you want, I can change it

apps/eodh-fe/tailwind.config.js
```diff
@@ -8,7 +9,14 @@ module.exports = {
     ...createGlobPatternsForDependencies(__dirname),
   ],
   theme: {
-    extend: {},
+    extend: {
+      colors: {
+        ...themeConfig.colors.light,
```
mlu-ssoft, line 14: there should be two themes, eg dark and light
you should import one of them (instead of importing specific properties, but whole object with whole configuration)

libs/shared/theme/src/lib/colors.ts
```diff
@@ -0,0 +1,126 @@
+const colors = {
+  light: {
```
mlu-ssoft, line 2: those should be separate themes, we should have dark/light theme with its own configuration

libs/shared/theme/src/lib/font-size.ts
```diff
@@ -0,0 +1,11 @@
+const fontSize = {
+  'small-semibold': ['8px', { lineHeight: '12px', fontWeight: '400' }],
+  'small-medium': ['9px', { lineHeight: '13.5px', fontWeight: '500' }],
+  'medium-bold': ['12px', { lineHeight: '18px', fontWeight: '700' }],
```
mlu-ssoft, line 4: shouldn't we use rem? maybe add todo to discuss it with tobi (his is on holidays right now)

libs/shared/theme/src/index.ts
```diff
@@ -0,0 +1,9 @@
+import colors from './lib/colors';
+import fontSize from './lib/font-size';
+
+const themeConfig = {
```
mlu-ssoft, line 4: there should be two themes exported: `darkTheme`, `lightTheme`

mlu-ssoft (changes requested):

mlu-ssoft (approved):

PR #16, opened 2024-07-21
feat: add possibility to draw aoi on a map

mlu-ssoft:


libs/map/ui-map/src/lib/aoi-layer.component.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    setDraw(draw);
+  }, [setDraw]);
+
+  return (
+    <div className='p-4 cursor-pointer hover:bg-[#4483FF] hover:text-[#FFFFFF]' onClick={drawPolygon}>
+      Polygon
+    </div>
+  );
+};
```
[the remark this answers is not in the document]
mlu-ssoft: as I mentioned on the daily, we will refactor this. We will create wrappers for OpenLayer and the we will need to rewrite this functionality. That's why I didn't put storybook or unit tests - feature at the end will look completly different coparing to what it is now

libs/map/ui-map/src/lib/aoi-layer/button.component.tsx
```diff
@@ -0,0 +1,18 @@
+import clsx from 'clsx';
+import { MouseEventHandler, PropsWithChildren } from 'react';
+
+export const DrawButton = ({
+  selected,
+  onClick,
+  children,
+}: PropsWithChildren<{ selected: boolean; onClick: MouseEventHandler<HTMLDivElement> }>) => {
```
KatarzynaPaszko, line 8: { selected: boolean; onClick: MouseEventHandler<HTMLDivElement> }
create interface for it

libs/map/ui-map/src/lib/aoi-layer/button.component.tsx
```diff
@@ -0,0 +1,18 @@
...
+  const className = selected
+    ? 'bg-primary text-primary-contrastText [&>svg]:fill-primary-contrastText'
+    : 'hover:bg-primary hover:text-primary-contrastText [&>svg]:fill-text [&>svg]:hover:fill-primary-contrastText';
+
+  return (
+    <div className={clsx('p-4 cursor-pointer', className)} onClick={onClick}>
+      {children}
+    </div>
```
KatarzynaPaszko, line 16: I think it should be `<button/> `tag

libs/map/ui-map/src/lib/aoi-layer/use-aoi-layer.hook.tsx
```diff
@@ -0,0 +1,87 @@
...
+import Feature from 'ol/Feature';
+import { Draw, Modify } from 'ol/interaction.js';
+import { DrawEvent } from 'ol/interaction/Draw.js';
+import VectorLayer from 'ol/layer/Vector';
+import VectorSource from 'ol/source/Vector';
+import { useContext, useEffect, useState } from 'react';
+
+import { MapContext } from '../map.component';
+
+export type TDraw = { draw: Draw; type: 'rectangle' | 'polygon' | 'circle' };
```
KatarzynaPaszko, line 11: In general, it is better to use interfaces over types whenever it is possible, as interfaces are cached therefore it works faster in ide

libs/map/ui-map/src/lib/aoi-layer/use-aoi-layer.hook.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export const useAioLayer = ({
+  draw,
+  setDraw,
+}: {
+  draw: TDraw | undefined;
+  setDraw: (draw: TDraw | undefined) => void;
+}) => {
```
KatarzynaPaszko, line 19: please add interface for it

libs/map/ui-map/src/lib/map.component.tsx
```diff
@@ -30,15 +39,15 @@ export const MapWrapper = ({ children }: PropsWithChildren) => {
       source: new OSM(),
     });
 
-    const map = new OlMap({
+    const olMap = new OlMap({
       layers: [osmLayer],
       view: new OlView({
-        center: [0, 0],
-        zoom: 0,
+        center: fromLonLat([-0.118092, 51.509865]),
```
KatarzynaPaszko, line 45: What are those values? Should we create cons with them, with name that would explain its value?

libs/map/ui-map/src/lib/aoi-layer/button.component.tsx
mlu-ssoft: why? what's benefit of this?

libs/map/ui-map/src/lib/aoi-layer/use-aoi-layer.hook.tsx
mlu-ssoft: it depends. yep, interfaces are faster, but it make difference only in very large codebases
and interfaces has a drawback: you can accidentally extend them
I preffer approach where I will have less errors (eg. I will get an error if I will try to extend type which I shouldn't). And types are very good for such use cases. If we need to have possibility to extend something, interfaces are better

KatarzynaPaszko (approved):

PR #24, opened 2024-08-05
feat: add left panel

mlu-ssoft:
- add left panel
- add logo
- fix top bar
- add search layout


apps/eodh-fe/src/app/layout/logo.component.tsx
KatarzynaPaszko, line 1: I think we should add logo to libs/shared, as it might be needed in other places as well in the future

libs/shared/theme/src/index.ts
```diff
@@ -30,5 +33,8 @@ const darkThemeConfig = {
   textShadow: {
     ...shadow,
   },
+  spacing: {
+    '6.5': '1.625rem',
+  },
```
KatarzynaPaszko, line 38: As some properties are same, I would consolidate it to avoid situation that we would miss adding something to some of the theme. 

```
const sharedThemeConfig = {
  fontSize: {
    ...fontSize,
  },
  boxShadow: {
    ...shadow,
  },
  textShadow: {
    ...shadow,
  },
  spacing: {
    '6.5': '1.625rem',
  },
};

const lightThemeConfig = {
  colors: {
    ...colors.light,
  },
  ...sharedThemeConfig,
};

const darkThemeConfig = {
  colors: {
    ...colors.dark,
  },
  ...sharedThemeConfig,
};
```

apps/eodh-fe/src/app/layout/search.component.tsx
KatarzynaPaszko, line 1: Why it is called SearchLayout?

apps/eodh-fe/src/app/layout/left-menu.component.tsx
KatarzynaPaszko, line 1: Please add hiding functionality to this component. Also, I think we should move it to libs/shared

KatarzynaPaszko (commented):
Panel is missing functionality to hide it, that will be crucial for tablet view. 
Screenshot 2024-08-07 at 10 20 26
Screenshot 2024-08-07 at 10 20 07

apps/eodh-fe/src/app/layout/logo.component.tsx
mlu-ssoft: then we will move it. For now it is not needed in other places. ATM we don't know where and in which places it will be used. I prefer do not complicate if not needed

apps/eodh-fe/src/app/layout/search.component.tsx
mlu-ssoft: because it is search, in the future we will have AC so maybe it will be different layout (with separate map instance)

apps/eodh-fe/src/app/layout/left-menu.component.tsx
mlu-ssoft: 1. this is layout specific, we can have different layouts and those layouts may vary between apps. Layout (and complex layout logic) may live in the app.
2. hidding functionality should be done in separate task as it requires creating custom controls for zoom on the map

libs/shared/theme/src/index.ts
mlu-ssoft: done

KatarzynaPaszko (approved):

PR #31, opened 2024-08-22
UKRIW-156 - date picker

KatarzynaPaszko:
Ticket: 
https://ukri-spyrosoft.atlassian.net/jira/software/projects/UKRIW/boards/1?assignee=5e4e41887ece1a0c9992c39a&selectedIssue=UKRIW-156

Description: Added date picker component


libs/shared/design-system/src/lib/icon/icon.tsx
```diff
@@ -11,7 +11,7 @@ export const Icon = ({ name = 'Bolt', width, height, className }: IIconProps) =>
   const IconComponent = IconsNames[name];
 
   return (
-    <div className={`min-w-6 flex justify-center items-center ${className}`}>
+    <div className={`${className} flex justify-center items-center`}>
```
mlu-ssoft, line 14: I think className should be at the end - it will allow us rewrite some styles if needed

libs/shared/design-system/src/lib/icon/icons/calendar.tsx
```diff
@@ -0,0 +1,12 @@
+import IIconProps from './icon.type';
+
+const Calendar = ({ width = 5, height = 8 }: IIconProps) => (
```
mlu-ssoft, line 3: width and height should be 24, double check that, but I think you copied inside icon (without bbox)

libs/shared/design-system/src/lib/input/date-input/date-input.tsx
```diff
@@ -0,0 +1,40 @@
...
+  value?: string;
+  onChange?: (value: string) => void;
+  className?: string;
+  error?: string;
+}
+
+export const DateInput: React.FC<IDateInputProps> = ({ value, onChange, className, error }) => {
+  const [inputValue, setInputValue] = useState(value || '');
+
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```
mlu-ssoft, line 19: add useCallback

mlu-ssoft (changes requested):

libs/shared/design-system/src/lib/icon/icons/calendar.tsx
KatarzynaPaszko: On next line you have bbox `viewBox='0 0 24 24' `

libs/shared/design-system/src/lib/icon/icon.tsx
KatarzynaPaszko: I did it as an experiment to check if the order matter, and it doesn't. We should investigate it, because right now even if className is at the end, it sometimes not overwrite styling of component

mlu-ssoft (approved):

PR #32, opened 2024-08-27
UKRIW-18 - data range in search module

KatarzynaPaszko:
Ticket: https://ukri-spyrosoft.atlassian.net/jira/software/projects/UKRIW/boards/1?assignee=5e4e41887ece1a0c9992c39a&selectedIssue=UKRIW-17


libs/map/ui-search-panel/src/lib/date-range-picker/date-range-picker.component.tsx
```diff
@@ -0,0 +1,94 @@
+import { DateInput, Icon, Text } from '@ukri/shared/design-system';
+import { useCallback, useState } from 'react';
+
+import { styles } from './date-range-picker.styles';
+
+const minDate = '1972-01-01';
+
+export const DateRangePicker = () => {
```
mlu-ssoft, line 8: pass min/max date as parameter

libs/map/ui-search-panel/src/lib/date-range-picker/date-range-picker.component.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const today = new Date();
+  const formattedToday = today.toISOString().split('T')[0];
+
+  const oneMonthAgo = new Date();
```
mlu-ssoft, line 12: move date callculation into hook

libs/map/ui-search-panel/src/lib/date-range-picker/date-range-picker.component.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const formattedOneMonthAgo = oneMonthAgo.toISOString().split('T')[0];
+
+  const [isOpen, setIsOpen] = useState(true);
+  const [fromDate, setFromDate] = useState<string | undefined>(formattedOneMonthAgo);
+  const [toDate, setToDate] = useState<string | undefined>(formattedToday);
+
+  const handleFromDateChange = useCallback(
+    (date: string) => {
+      setFromDate(date);
+      if (toDate && date > toDate) {
```
mlu-ssoft, line 23: move to separate compare function. you are assuming that dates are numbers, but it doesn't have be like that

libs/map/ui-search-panel/src/lib/date-range-picker/date-range-picker.component.tsx
```diff
@@ -0,0 +1,94 @@
...
+        setToDate(undefined);
+      }
+    },
+    [toDate]
+  );
+
+  const handleToDateChange = useCallback(
+    (date: string) => {
+      setToDate(date);
+      if (fromDate && date < fromDate) {
```
mlu-ssoft, line 33: same here, move to function

libs/map/ui-search-panel/src/lib/date-range-picker/date-range-picker.component.tsx
```diff
@@ -0,0 +1,94 @@
...
+        setFromDate(undefined);
+      }
+    },
+    [fromDate]
+  );
+
+  const toggleOpen = () => {
```
mlu-ssoft, line 40: add useCallback

libs/map/ui-search-panel/src/lib/search-panel.component.tsx
```diff
@@ -1,5 +1,18 @@
...
 export const SearchPanel = () => {
-  return <Tree />;
+  return (
+    <div className='flex flex-col h-full'>
+      <div className='flex-1 overflow-y-auto'>
+        <Tree />
+      </div>
+      <div className='mt-auto shadow-data-range-picker p-4'>
+        <DateRangePicker />
+        <Button text='Search' className='w-full flex justify-center mt-0' size='large' />
```
mlu-ssoft, line 14: add translation key for `Seach`

libs/shared/design-system/src/lib/input/date-input/date-input.tsx
```diff
@@ -27,14 +29,16 @@ export const DateInput: React.FC<IDateInputProps> = ({ value, onChange, classNam
   );
 
   return (
-    <div>
+    <div className=''>
```
mlu-ssoft, line 32: remove

mlu-ssoft (changes requested):

libs/map/ui-search-panel/src/lib/date-range-picker/date-range-picker.component.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+import { styles } from './date-range-picker.styles';
+import useDateCalculations from './dates-calculations.hook';
+
+interface IDateRangePickerProps {
+  minDate: string;
+  maxDate: string;
+}
+
+const isDateAfterToDate = (date: string, toDate?: string) => {
```
mlu-ssoft, line 12: maybe just `isDateAfter(date1, date2)` (or name date1/date2 params differently)?

libs/map/ui-search-panel/src/lib/date-range-picker/date-range-picker.component.tsx
```diff
@@ -0,0 +1,101 @@
...
+  return toDate && date > toDate;
+};
+
+const isDateBeforeFromDate = (date: string, fromDate?: string) => {
```
mlu-ssoft, line 16: maybe just `isDateBefore(date1, date2)` (or name date1/date2 params differently)?

libs/map/ui-search-panel/src/lib/date-range-picker/dates-calculations.hook.tsx
```diff
@@ -0,0 +1,16 @@
+import { useMemo } from 'react';
+
+const useDateCalculations = () => {
```
mlu-ssoft, line 3: I would split this into two functions:
- useDate which will return two dates: today, oneMonthAgo|
- formatDate which will format date

libs/map/ui-search-panel/src/lib/search-panel.component.tsx
```diff
@@ -1,5 +1,22 @@
...
 export const SearchPanel = () => {
-  return <Tree />;
+  const { formattedToday } = useDateCalculations();
+  return (
+    <div className='flex flex-col h-full'>
+      <div className='flex-1 overflow-y-auto'>
+        <Tree />
+      </div>
+      <div className='mt-auto shadow-data-range-picker p-4'>
+        <DateRangePicker maxDate={formattedToday} minDate={minDate} />
```
mlu-ssoft, line 17: why minDate isn't formatted as formattedToday date?

mlu-ssoft (changes requested):

libs/map/ui-search-panel/src/lib/date-range-picker/date-range-picker.component.tsx
KatarzynaPaszko: hmm, I think now name of the function and name of it parameters explains very good what this function does. 
is**Date**After`ToDate` = (**date**: string, `toDate`?: string)

If I will update it as per your suggestion, then you need to go back to function and analyze it which parameter should go first and which one should be second.

libs/map/ui-search-panel/src/lib/date-range-picker/date-range-picker.component.tsx
KatarzynaPaszko: Same as above

libs/map/ui-search-panel/src/lib/search-panel.component.tsx
KatarzynaPaszko: minDate is passed as a parameter. It is formatted already

libs/map/ui-search-panel/src/lib/date-range-picker/dates-calculations.hook.tsx
KatarzynaPaszko: Then we will need to import 2 hooks in every places, instead of one. And there is no case that we will use one, but not the other

mlu-ssoft: add then todo to split this function in the future. we will need one function for date formatting

mlu-ssoft (approved):

mlu-ssoft (approved):

PR #34, opened 2024-08-30
UKRIW-335 - display results on left panel

KatarzynaPaszko:
Ticket Acs: https://ukri-spyrosoft.atlassian.net/browse/UKRIW-62


libs/shared/design-system/src/lib/icon/icons/satellite-alt.tsx
```diff
@@ -0,0 +1,16 @@
+import IIconProps from './icon.type';
+
+const SatelliteAlt = ({ width = 48, height = 48 }: IIconProps) => (
+  <svg width={width} height={height} viewBox='0 0 48 48' fill='currentColor' xmlns='http://www.w3.org/2000/svg'>
```
KatarzynaPaszko, line 4: SateliteAlt is 48x48 by Toby request

apps/eodh-fe/public/assets/i18n/en.json
```diff
@@ -38,7 +38,18 @@
...
         "COMMON": "Something went wrong, please try again"
+      },
+      "SERVER_ERROR": {
+        "TITLE": "Server error",
+        "MESSAGE": "There was a problem retrieving your search results from the server. Please update your search parameters and try again."
+      },
+      "NO_RESULTS": {
+        "TITLE": " Sorry, your search returned 
no results.",
+        "MESSAGE": " Please refine your search criteria. Try a broader date range, adjust Data Set settings or consider choosing multiple Data Sets."
```
mlu-ssoft, line 48: remove empty space at beggining of the translation

libs/map/ui-results-panel/src/lib/results-list/results-list.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <div className='thumbnail-list'>
+      {results.map((thumbnail, index) => (
+        <Thumbnail
+          className='mb-4'
+          key={index}
+          {...thumbnail}
+          selected={selectedThumbnailIndex === index}
+          onSelected={() => handleThumbnailSelect(index)}
+          // actual functions to be added in future
```
mlu-ssoft, line 24: add todo

libs/map/ui-results-panel/src/lib/results-list/results-list.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { useState } from 'react';
+
+export interface IResultsListProps {
+  results: IThumbnailProps[];
+}
+
+export const ResultsList = ({ results }: IResultsListProps) => {
+  const [selectedThumbnailIndex, setSelectedThumbnailIndex] = useState<number | null>(null);
+
+  const handleThumbnailSelect = (index: number) => {
```
mlu-ssoft, line 11: add useCallback

libs/map/ui-results-panel/src/lib/results-list/results-list.stories.tsx
```diff
@@ -0,0 +1,43 @@
+import type { Meta } from '@storybook/react';
+
+import { ResultsList } from './results-list.component';
+
+const singleElementMock = {
+  imageUrl: '/assets/images/imageSample2.png', // image is added in the assets folder in storybook-host
```
mlu-ssoft, line 6: maybe instead of adding comment use proper variable name? eg: `const imageUrlStoredInPublicAssetsInStorybookHost = '...'`

libs/map/ui-results-panel/src/lib/results-panel.component.tsx
```diff
@@ -0,0 +1,53 @@
...
+    }
+  | {
+      status: 'success';
+      data: [] | IThumbnailProps[];
+    };
+
+type TResultsPanelProps = TResultsStateProps & IBaseResultsPanelProps;
+
+export const ResultsPanel = (props: TResultsPanelProps) => {
+  if (props.status === 'success') {
```
mlu-ssoft, line 26: you can use one if condition

libs/map/ui-results-panel/src/lib/results-panel.component.tsx
```diff
@@ -0,0 +1,53 @@
...
+          message='GLOBAL.ERRORS.NO_RESULTS.MESSAGE'
+          buttonText='GLOBAL.NAVIGATION.RETURN_TO_SERCH'
+          buttonOnClick={props.onReturn}
+        />
+      );
+    }
+    return <ResultsList results={props.data} />;
+  }
+
+  if (props.status === 'error') {
```
mlu-ssoft, line 41: consider using `switch-case`

libs/map/ui-results-panel/src/lib/results-panel.component.tsx
```diff
@@ -0,0 +1,53 @@
+import { ApiError } from '@ukri/shared/design-system';
+import { type IThumbnailProps } from '@ukri/shared/design-system';
+
+import { ResultsList } from './results-list/results-list.component';
+
+interface IBaseResultsPanelProps {
+  onReturn: () => void;
```
mlu-ssoft, line 7: maybe `onBack`?

libs/map/ui-results-panel/src/lib/results-panel.stories.tsx
```diff
@@ -0,0 +1,50 @@
+import type { Meta } from '@storybook/react';
+
+import { ResultsPanel } from './results-panel.component';
+
+const singleElementMock = {
+  imageUrl: '/assets/images/imageSample2.png', // image is added in the assets folder in storybook-host
```
mlu-ssoft, line 6: maybe instead of adding comment use proper variable name? eg: `const imageUrlStoredInPublicAssetsInStorybookHost = '...'`

libs/map/ui-results-panel/src/lib/results-panel.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+  title: 'libs/map/ui-results-panel/ResultsPanel',
+};
+export default meta;
+
+export const ResultsPanelSample = {
+  args: {
+    status: 'success',
+    data: results,
+    error: { message: 'Results not found' },
+    onReturn: () => {
```
mlu-ssoft, line 46: you can add eslint-ignore rule for empty function in this case

libs/shared/design-system/src/lib/api-error/api-error.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+interface IApiErrorProps {
+  iconName?: TIconNames | null;
+  title: string;
+  message: string;
+  buttonText?: string;
+  buttonOnClick?: () => void;
+}
+
+export const ApiError = ({ iconName, title, message, buttonText, buttonOnClick }: IApiErrorProps) => {
```
mlu-ssoft, line 13: maybe just `Error`?

libs/shared/design-system/src/lib/icon/icons/satellite-alt.tsx
mlu-ssoft: can we create component which will have 24x24 and make it bigger in places where it is needed? I would prefer to have all icons in the same size in Design System

libs/shared/design-system/src/lib/thumbnail/thumbnail.tsx
```diff
@@ -18,7 +18,10 @@ const ThumbnailInfo = ({ name, iconName }: IThumbnailInfoProps) => {
   );
 };
 
-interface IThumbnailProps {
+// Added for now, to be rempved in the foture once we will work on comparison functionality
```
mlu-ssoft, line 21: add todo. if something should be done in the future use todo. it will be easier to search in the project, eg search for // todo

libs/shared/design-system/src/lib/thumbnail/thumbnail.tsx
```diff
@@ -41,24 +45,36 @@ export const Thumbnail = ({
...
+  const handleSelectItem = useCallback(() => {
     setIsSelected(!isSelected);
     if (onSelected) {
       onSelected();
     }
-  };
+  }, [isSelected, onSelected]);
+
+  const handleCompareClick = useCallback(() => {
+    if (isAddedForComparison === false && onAddToCompare) {
```
mlu-ssoft, line 64: it can be simplified: `if (!isAddedForComparison && onAddToCompare) {`

libs/shared/design-system/src/lib/thumbnail/thumbnail.tsx
```diff
@@ -41,24 +45,36 @@ export const Thumbnail = ({
...
+      setIsAddedForComparison(true);
+      onAddToCompare();
+    } else if (isAddedForComparison === true && onRemoveFromCompare) {
```
mlu-ssoft, line 67: it can bi simplified: `else if (isAddedForComparison && onRemoveFromCompare) {`

libs/shared/design-system/src/lib/thumbnail/thumbnail.tsx
```diff
@@ -41,24 +45,36 @@ export const Thumbnail = ({
...
+      setIsAddedForComparison(false);
```
mlu-ssoft, line 68: why this is invoked only if there is `onRemoveFromCompare` callback?

libs/shared/design-system/src/lib/thumbnail/thumbnail.tsx
```diff
@@ -41,24 +45,36 @@ export const Thumbnail = ({
...
     setIsSelected(!isSelected);
     if (onSelected) {
       onSelected();
     }
-  };
+  }, [isSelected, onSelected]);
+
+  const handleCompareClick = useCallback(() => {
+    if (isAddedForComparison === false && onAddToCompare) {
+      setIsAddedForComparison(true);
```
mlu-ssoft, line 65: why this is invoked only if there is `onAddToCompare` callback?

libs/shared/design-system/src/lib/thumbnail/thumbnail.tsx
```diff
@@ -41,24 +45,36 @@ export const Thumbnail = ({
   gridCode,
   selected,
   onSelected,
-  addedForComparison,
+  addedForComparison = false,
```
mlu-ssoft, line 48: this name suggest for what it will be used. This component should be more generic. Maybe `showCompareButton`?

libs/shared/design-system/src/lib/thumbnail/thumbnail.tsx
```diff
@@ -28,8 +31,9 @@ interface IThumbnailProps {
   onSelected?: () => void;
   selected?: boolean;
   addedForComparison?: boolean;
-  onAddToCompare?: () => void;
-  onRemoveFromCompare?: () => void;
+  onAddToCompare: () => void;
```
mlu-ssoft, line 34: you could use one callback: `onCompareToggle` or just `onCompare` with `true/false` or `mode: add/remove` as a parameter

libs/shared/design-system/src/lib/api-error/api-error.tsx
```diff
@@ -0,0 +1,24 @@
+import { Button } from '../button/button';
+import { Icon, type TIconNames } from '../icon/icon';
+import { Text } from '../text/text';
+
+interface IApiErrorProps {
+  iconName?: TIconNames | null;
+  title: string;
+  message: string;
+  buttonText?: string;
+  buttonOnClick?: () => void;
```
mlu-ssoft, line 10: maybe just `onClick`?

libs/shared/design-system/src/lib/api-error/api-error.tsx
```diff
@@ -0,0 +1,24 @@
+import { Button } from '../button/button';
+import { Icon, type TIconNames } from '../icon/icon';
+import { Text } from '../text/text';
+
+interface IApiErrorProps {
+  iconName?: TIconNames | null;
+  title: string;
+  message: string;
+  buttonText?: string;
```
mlu-ssoft, line 9: 1. type should be ParseKeys from i18n
2. maybe `actionText`? there is title and message, those don't suggest where are used, eg there is no divTitle or divMessage

libs/shared/design-system/src/lib/api-error/api-error.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+export default meta;
+
+export const SampleApiError = {
+  args: {
+    iconName: null,
+    title: 'Sorry, your search returned no results.',
+    message:
+      'Please refine your search criteria. Try a broader date range, adjust Data Set settings or consider choosing multiple Data Sets.',
+    buttonText: 'Return to the search menu',
+    buttonOnClick: () => {
```
mlu-ssoft, line 18: you could add eslint disable rule

mlu-ssoft (changes requested):

libs/shared/design-system/src/lib/icon/icons/satellite-alt.tsx
KatarzynaPaszko: I will update height and width, but I can't mess up with viewBox, coz then icon is broken.

libs/shared/design-system/src/lib/thumbnail/thumbnail.tsx
KatarzynaPaszko: because we want to make sure that it is not just visual change on Thumbnail/ResultItem. We want to make sure that this Item is indeed being proces by function and added for comparison.

libs/shared/design-system/src/lib/thumbnail/thumbnail.tsx
KatarzynaPaszko: Same as above

libs/shared/design-system/src/lib/api-error/api-error.tsx
KatarzynaPaszko: updated to ctaOnClick

libs/shared/design-system/src/lib/api-error/api-error.tsx
KatarzynaPaszko: updated to ctaText

libs/shared/design-system/src/lib/icon/icons/satellite-alt.tsx
mlu-ssoft: you can ask Tobi to add this icon to icons and then download it

mlu-ssoft (changes requested):

libs/map/ui-results-panel/src/lib/results-list/results-list.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const ResultsList = ({ results }: IResultsListProps) => {
+  const [selectedThumbnailIndex, setSelectedThumbnailIndex] = useState<number | null>(null);
+
+  const handleThumbnailSelect = useCallback((index: number) => {
+    setSelectedThumbnailIndex(index);
+  }, []);
+
+  return (
+    <div className='thumbnail-list'>
```
mlu-ssoft, line 16: why css class is here?

libs/map/ui-results-panel/src/lib/results-list/results-list.component.tsx
```diff
@@ -0,0 +1,35 @@
+import { type IThumbnailProps, Thumbnail } from '@ukri/shared/design-system';
+import { useCallback, useState } from 'react';
+
+export interface IResultsListProps {
+  results: IThumbnailProps[];
+}
+
+export const ResultsList = ({ results }: IResultsListProps) => {
+  const [selectedThumbnailIndex, setSelectedThumbnailIndex] = useState<number | null>(null);
```
mlu-ssoft, line 9: consider using ID, I think each result should have ID property
index may change (while rerendering) so I think we should at least reset index then results property changes

libs/map/ui-results-panel/src/lib/results-panel.component.tsx
```diff
@@ -0,0 +1,56 @@
...
+  | {
+      status: 'success';
+      data: [] | IThumbnailProps[];
+    };
+
+type TResultsPanelProps = TResultsStateProps & IBaseResultsPanelProps;
+
+export const ResultsPanel = (props: TResultsPanelProps) => {
+  switch (props.status) {
+    case 'success':
```
mlu-ssoft, line 27: add braces, it is easly to miss that there is return condition below

libs/map/ui-results-panel/src/lib/results-panel.component.tsx
```diff
@@ -0,0 +1,56 @@
...
+            title='GLOBAL.ERRORS.NO_RESULTS.TITLE'
+            message='GLOBAL.ERRORS.NO_RESULTS.MESSAGE'
+            ctaText='GLOBAL.NAVIGATION.RETURN_TO_SERCH'
+            ctaOnClick={props.onBack}
+          />
+        );
+      }
+      return <ResultsList results={props.data} />;
+
+    case 'error':
```
mlu-ssoft, line 41: add braces

libs/map/ui-results-panel/src/lib/results-panel.component.tsx
```diff
@@ -0,0 +1,56 @@
...
+          title='GLOBAL.ERRORS.SERVER_ERROR.TITLE'
+          message='GLOBAL.ERRORS.SERVER_ERROR.MESSAGE'
+          ctaText='GLOBAL.NAVIGATION.RETURN_TO_SERCH'
+          ctaOnClick={props.onBack}
+        />
+      );
+
+    case 'loading':
+    case 'idle':
+    default:
```
mlu-ssoft, line 53: add braces

libs/shared/design-system/src/lib/thumbnail/thumbnail.tsx
mlu-ssoft, line 34: I would create two types for IThumbnailProps, one without onAddToCompare and onRemoveFromCompare functions and second one with those functions. Then it will be clear that you have to pass two of those functions (or none of them)

mlu-ssoft (changes requested):

libs/map/ui-results-panel/src/lib/results-list/results-list.component.tsx
KatarzynaPaszko: I was testing something, and forgot to remove. Its removed now

mlu-ssoft (approved):

PR #37, opened 2024-09-03
UKRIW-14 - display footprints on map

KatarzynaPaszko:
Ticket requirements: https://ukri-spyrosoft.atlassian.net/browse/UKRIW-14


libs/map/ui-map/src/lib/footprints-layer/use-footprint-layer.hook.tsx
```diff
@@ -0,0 +1,105 @@
...
+    const vectorSource = new VectorSource({
+      features: new GeoJSON().readFeatures(geojsonObject, {
+        featureProjection: map.getView().getProjection(),
+      }),
+    });
+
+    const newVectorLayer = new VectorLayer({
+      source: vectorSource,
+      style: defaultStyle,
+      zIndex: 2,
```
mlu-ssoft, line 57: add this as parameter (or move to variable, eg layersOrder)

libs/map/ui-map/src/lib/aoi-layer/use-aoi-layer.hook.tsx
```diff
@@ -21,6 +21,7 @@ export const useAioLayer = ({ draw, setDraw }: TAoiLayer) => {
     const vectorSource = new VectorSource({ wrapX: false });
     const vector = new VectorLayer({
       source: vectorSource,
+      zIndex: 1,
```
mlu-ssoft, line 24: add this as parameter (or move to variable, eg layersOrder)

libs/map/ui-map/src/lib/footprints-layer/use-footprint-layer.hook.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { Fill, Stroke, Style } from 'ol/style';
+import { useCallback, useContext, useEffect, useState } from 'react';
+
+import { MapContext } from '../map.component';
+import { type IFeatureCollection } from './geo-json.type';
+
+const defaultStyle = new Style({
+  fill: new Fill({
+    // Fill do not support opacity, so we need to use rgba here
+    color: 'rgba(68, 131, 255, 0.2)',
```
mlu-ssoft, line 17: add todo to check if it is possible to get styles from tailwind

libs/map/ui-map/src/lib/footprints-layer/use-footprint-layer.hook.tsx
```diff
@@ -0,0 +1,105 @@
...
+    });
+
+    const newVectorLayer = new VectorLayer({
+      source: vectorSource,
+      style: defaultStyle,
+      zIndex: 2,
+      visible: isVisible,
+    });
+
+    setVectorLayer(newVectorLayer);
```
mlu-ssoft, line 61: shouldn't it be done at the end of this hook?

libs/map/ui-map/src/lib/footprints-layer/use-footprint-layer.hook.tsx
```diff
@@ -0,0 +1,105 @@
...
+    map.addLayer(newVectorLayer);
+
+    const selectHover = new Select({
```
mlu-ssoft, line 64: consider move to another hook (split into vector creation and adding interaction)

libs/map/ui-map/src/lib/footprints-layer/use-footprint-layer.hook.tsx
```diff
@@ -0,0 +1,105 @@
...
+    },
+    [vectorLayer]
+  );
+
+  const toggleVisibility = useCallback(
+    (visible?: boolean) => {
+      if (vectorLayer) {
+        const newVisibility = visible !== undefined ? visible : !vectorLayer.getVisible();
+        vectorLayer.setVisible(newVisibility);
+        setIsVisible(newVisibility);
```
mlu-ssoft, line 98: is it needed to have additional property to trigger visibility?
Right now it works as follow:
1. set visibility on map
2. change property in use state
3. property is changed, view layer is removed
4. new layer is created

I think it will be enough to use `vectorLayer` API

mlu-ssoft (changes requested):

libs/map/ui-map/src/lib/footprints-layer/use-footprint-layer.hook.tsx
KatarzynaPaszko: hmm, there wont be a case, that you would want to add those interactions to another layer, nor that you would create footprints without this interactions, so I do not think dividing into 2 hooks make sense in this particular case.

libs/map/ui-map/src/lib/footprints-layer/use-footprint-layer.hook.tsx
KatarzynaPaszko: what difference would it make in this case?

libs/map/ui-map/src/lib/footprints-layer/use-footprint-layer.hook.tsx
KatarzynaPaszko: I have created toggleVisibility so we can hide footprints once displaying image, and then once we are hiding image we can again show footprints. As per yesterday conversation, flow might change, so we might not need it in the future, but I left it for now, as we do not know yet if it will be needed or not

libs/map/ui-map/src/lib/footprints-layer/use-footprint-layer.hook.tsx
mlu-ssoft: it may makre difference from reading the code my first question is: what will happens next. To answer that I have to go to another hooks (to check if those hooks won't trigger anything in the hook that I read the code right now)

mlu-ssoft (changes requested):

libs/map/ui-map/src/lib/footprints-layer/use-footprint-layer.hook.tsx
```diff
@@ -0,0 +1,106 @@
...
+      features: new GeoJSON().readFeatures(geojsonObject, {
+        featureProjection: map.getView().getProjection(),
+      }),
+    });
+
+    const newVectorLayer = new VectorLayer({
+      source: vectorSource,
+      style: defaultStyle,
+      zIndex: footprintsLayerZindex,
+      // visible: isVisible,
```
mlu-ssoft, line 59: remove

mlu-ssoft (changes requested):

mlu-ssoft (approved):

PR #43, opened 2024-09-06
chore: display true color image on the map

mlu-ssoft:
Display true color image on the map:
- add store (for now it is as it is, it may change in the future when Anti Coruption Layer will be done for OpenLayers). I add it because there was conflict between the same layers (`ui-map` an `ui-results-view`, I couldn't use hook from ui-map in `ui-results-view`)
- add new layer (StacLayer)

Potential issues:
- new layer isn't visible as other Footprints make dificult to view the image

Tickets:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-73


libs/map/ui-map/src/lib/stac-layer/use-stac-layer.hook.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    const newStacLayer = new STAC({
+      url,
+      zIndex: stacLayerZindex,
+    });
+
+    newStacLayer.addEventListener('sourceready', () => {
+      const view = map.getView();
+      const extent = newStacLayer.getExtent();
+      if (extent) {
```
mlu-ssoft: add one empty line for readibility reason

libs/map/ui-map/src/lib/stac-layer/use-stac-layer.hook.tsx
```diff
@@ -0,0 +1,65 @@
...
+        view.fit(extent);
+        const zoom = view.getZoom();
+        if (zoom) {
```
mlu-ssoft: add one empty line for readibility reason

libs/map/ui-map/src/lib/stac-layer/use-stac-layer.hook.tsx
```diff
@@ -0,0 +1,65 @@
...
+    if (!map || !url) {
+      return;
+    }
+
+    const newStacLayer = new STAC({
+      url,
+      zIndex: stacLayerZindex,
+    });
+
+    newStacLayer.addEventListener('sourceready', () => {
```
mlu-ssoft: why use of `addEventListener` not `on` method? double check if you don't need to remove this event then (there is no `off` method, but there should be `removeEventListener` method

libs/shared/design-system/src/lib/result-item/result-item.tsx
```diff
@@ -130,7 +130,13 @@ export const ResultItem = ({
             onClick={handleCompareClick}
             className={`${hideCompareButton} ${isAddedForComparison ? '!text-error' : ''}`}
           />
-          <Button text='GLOBAL.DESIGN_SYSTEM.RESULT_ITEM.BUTTON' size='small' onClick={handleSelectItem} />
+          <Button
+            text={
+              selected ? 'GLOBAL.DESIGN_SYSTEM.RESULT_ITEM.BUTTON_HIDE' : 'GLOBAL.DESIGN_SYSTEM.RESULT_ITEM.BUTTON_SHOW'
+            }
+            size='small'
+            onClick={handleSelectItem}
```
mlu-ssoft, line 138: shouldn't it be `toggleItem` then? Right now its not only marks as selected but also change state from selected to unselected

apps/eodh-fe/src/app/layout/left-menu.component.tsx
```diff
@@ -2,7 +2,7 @@ import { SearchModePanel } from '@ukri/map/feature-search-mode-panel';
 
 export const LeftMenu = () => {
   return (
-    <div className='w-[360px] min-w-[360px] h-full bg-bright-main border-bright-mid border-r-[1px] overflow-scroll'>
+    <div className='w-[361px] min-w-[361px] h-full bg-bright-main border-bright-mid border-r-[1px] overflow-scroll'>
```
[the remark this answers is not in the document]
KatarzynaPaszko: Our designs seem to be pixel perfect. This component has border 1px, that on designs in inside, but in browsers is added outside. 

Once I added margin around `ResultsList`, then `"View"` button got squeezed, and do not look as on design anymore.

KatarzynaPaszko (approved):

apps/eodh-fe/src/app/layout/left-menu.component.tsx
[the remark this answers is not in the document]
mlu-ssoft: maybe use box sizing? I do not like use values like 361 (to mach to pixel), it look weird and in few years noone will remember why its 361 not 360. also I thing 360 value is used in top navigation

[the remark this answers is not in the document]
KatarzynaPaszko: I have added box sizing: content box

PR #46, opened 2024-09-09
feat: add toggle layers visibility button

mlu-ssoft:
- add toggle layers visibility button
- create store for Footprints
- move toggle visibility functionality into stores in data-access layer

Problems:
- there was problem with importing from `data-access-stac-catalog` in `data-access-map`. I disabled import (for now), lets see how stores will look like when Anti-Corruption Layer will be finished. Maybe this problem will be fixed then. If not then we have to think if we shoulnd't merge two data-access (`data-access-stac-catalog` and `data-access-map`) into one

Tasks:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-356


apps/eodh-fe/src/app/layout/default-layout.component.tsx
```diff
@@ -11,6 +11,7 @@ export const DefaultLayout = () => {
         <div className='flex w-full h-full overflow-hidden'>
           <LeftMenu />
           <TrueColorImageLayer />
+          <FootprintLayerComponent />
```
KatarzynaPaszko, line 14: No need to add `Component` to `FootprintLayerComponent`, simply `FootprintLayer` is enough, as in above `TrueColorImageLayer`

libs/map/data-access-map/src/lib/footprint.store.tsx
```diff
@@ -0,0 +1,37 @@
+import type {} from '@redux-devtools/extension';
+// eslint-disable-next-line @nx/enforce-module-boundaries
+import { TCollection } from '@ukri/map/data-access-stac-catalog';
```
KatarzynaPaszko, line 3: Add TODO comment, to look at it in the future and remove this disable

libs/map/ui-map/src/lib/footprints-layer/footprint-multiple-layers.hook.stories.tsx
```diff
@@ -1,26 +1,27 @@
...
 const GeoJsonLayerComponent = ({ resultItem }: { resultItem: TCollection }) => {
-  const { updateZindex, toggleVisibility } = useFootprintsLayer(resultItem);
+  const setCollection = useFootprintCollectionMutation();
+  const toggleVisibility = useToggleFootprintLayer();
+
+  useEffect(() => {
+    setCollection(resultItem);
+  }, [resultItem, setCollection]);
 
   return (
```
KatarzynaPaszko, line 23: Once testing this example in storybook, it does not work as expected. It creates just 1 footprint layer out of one mock, not 5 different layers from all different mocks.

KatarzynaPaszko (commented):
Looking at design 
https://www.figma.com/design/WDCJlvNjehFTkHVb4FSulP/UKRI-(original-bid-design)?node-id=4003-220761&t=gG9w9nklOmw4m2xA-4
by default eye icon should not be active, just after click (eye icon crossed out) it becomes active

KatarzynaPaszko (commented):
In this ticket (as you can read in technical notes) we were suppose to update look of drawing tools buttons as per this design: https://www.figma.com/design/WDCJlvNjehFTkHVb4FSulP/UKRI-(original-bid-design)?node-id=3969-193941&t=gG9w9nklOmw4m2xA-4

libs/map/ui-map/src/lib/footprints-layer/footprint-multiple-layers.hook.stories.tsx
mlu-ssoft: fixed

PR #59, opened 2024-09-23
UKRIW-80 -  (subtask UKRIW-365) action creator nodes component

KatarzynaPaszko:


libs/shared/design-system/src/lib/icon/icons/action-creator-arrow.tsx
```diff
@@ -0,0 +1,10 @@
+const ActionCreatorArrow = () => (
+  <svg xmlns='http://www.w3.org/2000/svg' width='8' height='29' viewBox='0 0 8 29' fill='none'>
+    <path
+      d='M4 0.333333C2.52724 0.333333 1.33333 1.52724 1.33333 3C1.33333 4.47276 2.52724 5.66667 4 5.66667C5.47276 5.66667 6.66667 4.47276 6.66667 3C6.66667 1.52724 5.47276 0.333333 4 0.333333ZM3.64645 28.3536C3.84171 28.5488 4.15829 28.5488 4.35355 28.3536L7.53553 25.1716C7.7308 24.9763 7.7308 24.6597 7.53553 24.4645C7.34027 24.2692 7.02369 24.2692 6.82843 24.4645L4 27.2929L1.17157 24.4645C0.97631 24.2692 0.659727 24.2692 0.464465 24.4645C0.269203 24.6597 0.269203 24.9763 0.464465 25.1716L3.64645 28.3536ZM3.5 3L3.5 28L4.5 28L4.5 3L3.5 3Z'
+      fill='#4483FF'
```
KatarzynaPaszko, line 5: I added color for now, as this is temporary arrow that will be replaced by ReactFlow framework arrows

libs/map/feature-action-creator-panel/src/lib/content/workflow/pseudo-input.component.tsx
```diff
@@ -0,0 +1,47 @@
...
+};
+
+interface IPseudoInputProps {
+  value?: string;
+  onChange?: (value: string) => void;
+  className?: string;
+  iconName?: 'Polygon' | 'Circle' | 'Satellite' | 'Square';
+}
+
+export const PseudoInput: React.FC<IPseudoInputProps> = ({ value, onChange, className, iconName = 'Satellite' }) => {
```
mlu-ssoft, line 21: this should be normal input. In the future we will use react hook form to validate data in Workflow tab

libs/shared/design-system/src/lib/icon/icons/action-creator-arrow.tsx
mlu-ssoft: ok

libs/shared/design-system/src/lib/select/select.tsx
```diff
@@ -1,76 +1,120 @@
...
 }
 
 export const Select = ({
   options,
   onChange,
   placeholder = 'GLOBAL.DESIGN_SYSTEM.SELECT.PLACEHOLDER',
   error,
+  className,
+  value,
 }: ISelectProps) => {
```
mlu-ssoft, line 32: This select should be not connected component (without internal state). This way we will be able to use it with react hook forms. If you need component with state then create two components - one basic, second one with state (eg. ConnectedSelect or something like that)

libs/map/feature-action-creator-panel/src/lib/content/workflow/function-input.component.tsx
```diff
@@ -0,0 +1,23 @@
...
+  { value: 'False colour (urban)', label: 'False colour (urban)' },
+  { value: 'Moisture index', label: 'Moisture index' },
+  { value: 'SWIR', label: 'SWIR' },
+  { value: 'NDWI', label: 'NDWI' },
+  { value: 'NDSI', label: 'NDSI' },
+];
+
+interface IFunctionSelectInputProps {
+  visible?: boolean;
+  onChange: (value: string) => void;
```
mlu-ssoft, line 14: I would this optional

libs/map/feature-action-creator-panel/src/lib/content/workflow/workflow.component.tsx
```diff
@@ -0,0 +1,45 @@
...
+  disabled: true,
+  selected: false,
+};
+
+export const Workflow = () => {
+  const showInputs = false;
+  return (
+    <section className='text-text-primary flex justify-center'>
+      <div>
+        <Node {...args} type='area'>
```
mlu-ssoft, line 19: I would create new component for each type: AreaComponent, DataSetsComponent, DateRangeComponent, FunctionComponent and here just display those components. Then each component will have its own logic

libs/map/feature-action-creator-panel/src/lib/content/workflow/workflow.component.tsx
```diff
@@ -0,0 +1,45 @@
...
+          {showInputs && <PseudoInput iconName='Polygon' value={'sample aoi'} />}
```
mlu-ssoft, line 20: can be just string, same in other `PseudoInput` components

mlu-ssoft (changes requested):

libs/shared/design-system/src/lib/select/select.tsx
KatarzynaPaszko: added comment to do in another ticket

libs/map/feature-action-creator-panel/src/lib/content/workflow/node-select.component.tsx
```diff
@@ -0,0 +1,32 @@
...
+      { value: 'NDVI', label: t(`${labelPath}.NDVI`) },
+      { value: 'False colour (urban)', label: t(`${labelPath}.FALSE_COLOUR`) },
+      { value: 'Moisture index', label: t(`${labelPath}.MOISTURE_INDEX`) },
+      { value: 'SWIR', label: t(`${labelPath}.SWIR`) },
+      { value: 'NDWI', label: t(`${labelPath}.NDWI`) },
+      { value: 'NDSI', label: t(`${labelPath}.NDSI`) },
+    ];
+  }, [t]);
+
+  const handleChange = (value?: string | null) => {
```
mlu-ssoft, line 25: add useCallback

libs/shared/design-system/src/lib/select/select.tsx
```diff
@@ -1,76 +1,121 @@
...
-      onChange(selectedOptionValue);
+      onChange(selectedOptionValue.value);
+      setIsOpen(true);
     },
     [onChange]
   );
 
+  useEffect(() => {
+    const handleClickOutside = (event: MouseEvent) => {
+      if (ref.current && !(ref.current as Node).contains(event.target as Node)) {
```
mlu-ssoft, line 54: maybe there should be different type for useRef?

mlu-ssoft (changes requested):

mlu-ssoft (approved):

libs/shared/design-system/src/lib/select/select.tsx
KatarzynaPaszko: Will be done in another ticket

PR #66, opened 2024-09-27
UKRIW-358 - cloud coverage bug fix

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-358

Fixed bug: Once cloud coverage is 0, Cloud coverage metadata is displayed wrongly

### DIFF
--- a/libs/shared/design-system/src/lib/result-item/result-item.tsx
+++ b/libs/shared/design-system/src/lib/result-item/result-item.tsx
@@ -1,3 +1,4 @@
+import isNumber from 'lodash/isNumber';
 import { useCallback, useMemo, useState } from 'react';
 
 import { Button } from '../button/button';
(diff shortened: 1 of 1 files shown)


libs/shared/design-system/src/lib/result-item/result-item.tsx
```diff
@@ -95,6 +95,10 @@ export const ResultItem = ({
     }
   }, [onAddToCompare, isAddedForComparison, onRemoveFromCompare]);
 
+  const cloudCoverageValue = useMemo(() => {
+    return typeof cloudCoverage === 'number' ? `${cloudCoverage.toFixed(2)}%` : cloudCoverage;
```
mlu-ssoft, line 99: use `isNumber` from lodash. Checking if it is number is more complex

mlu-ssoft (changes requested):

libs/shared/design-system/src/lib/result-item/result-item.tsx
KatarzynaPaszko: done

mlu-ssoft (approved):

PR #67, opened 2024-09-27
UKRIW-415 - fixed button active state for Safari browser

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-415
This pull request introduces several enhancements to the button component in the design system, focusing on adding an `isActive` state to handle active styling and improving the button's click and focus behavior especially on Safari browser.

### DIFF
--- a/libs/shared/design-system/src/lib/button/button.tsx
+++ b/libs/shared/design-system/src/lib/button/button.tsx
@@ -41,7 +41,6 @@ export const Button = ({
 }: IButtonProps) => {
   const buttonRef = useRef<HTMLButtonElement>(null);
   const { isActive, setIsActive } = useSafariFix({ buttonRef });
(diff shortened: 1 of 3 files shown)


libs/shared/design-system/src/lib/button/button.tsx
```diff
@@ -38,10 +39,13 @@ export const Button = ({
   disabled = false,
   type = 'button',
 }: IButtonProps) => {
-  const baseStyles = getBaseStyles(disabled, appearance);
+  const buttonRef = useRef<HTMLButtonElement>(null);
+  const { isActive, setIsActive } = useSafariFix({ buttonRef });
+  // const [isActive, setIsActive] = useState(false);
```
mlu-ssoft, line 44: remove

mlu-ssoft (approved):

PR #69, opened 2024-10-02
feat: add data store

mlu-ssoft:
- add data store
- add tests
- add query storage
- switch between query storage and memory storage

stories:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-384

### DIFF
--- a/libs/map/data-access-map/src/lib/data-store/data.model.ts
+++ b/libs/map/data-access-map/src/lib/data-store/data.model.ts
@@ -1,0 +1,25 @@
+import { defaultValues, TSearchDefaultValues } from './search.model';
+import { TSearchData } from './search.schema';
+
+export type TMode = 'search' | 'action-creator';
+export type TSchema = 'search' | 'action-creator';
+
+type TInitialData = TSearchData | TSearchDefaultValues | undefined;
+type TUpdateData = TSearchData | undefined;
+
+export type TDataStore<T = TInitialData, U = TUpdateData> = {
+  mode: TMode;
+  schema: TSchema;
+  data: U;
+  state: T;
+  updateData: (data: U) => void;
+  updateState: (data: T) => void;
+  toggleMode: () => void;
+};
+
+export const defaultState: Omit<TDataStore, 'updateData' | 'updateState' | 'toggleMode'> = {
+  mode: 'search',
+  schema: 'search',
+  state: defaultValues,
+  data: undefined,
+};
--- a/libs/map/data-access-map/src/lib/data-store/data.model.ts
+++ b/libs/map/data-access-map/src/lib/data-store/data.model.ts
@@ -1,11 +1,10 @@
+import { TSearchData } from './schema/search/search.schema';
 import { defaultValues, TSearchDefaultValues } from './search.model';
-import { TSearchData } from './search.schema';
-
 export type TMode = 'search' | 'action-creator';
 export type TSchema = 'search' | 'action-creator';
 
-type TInitialData = TSearchData | TSearchDefaultValues | undefined;
-type TUpdateData = TSearchData | undefined;
+export type TInitialData = TSearchData | TSearchDefaultValues | undefined;
+export type TUpdateData = TSearchData | undefined;
 
 export type TDataStore<T = TInitialData, U = TUpdateData> = {
   mode: TMode;
@@ -13,7 +12,7 @@ export type TDataStore<T = TInitialData, U = TUpdateData> = {
   data: U;
   state: T;
   updateData: (data: U) => void;
-  updateState: (data: T) => void;
+  updateState: (state: T) => void;
   toggleMode: () => void;
 };
 
--- a/libs/map/data-access-map/src/lib/data-store/search.model.ts
+++ b/libs/map/data-access-map/src/lib/data-store/search.model.ts
@@ -1,0 +1,150 @@
+import { createDateString, formatDate, TDateString } from '@ukri/shared/utils/date';
+import Geometry from 'ol/geom/Geometry';
+
+export type TSearchDefaultValues = {
+  dataSets: {
+    copernicus: {
+      enabled: boolean;
+      sentinel1: {
+        enabled: boolean;
+        expanded: boolean;
+        acquisitionMode: {
+          ew: boolean;
+          hh: boolean;
+          hh_hv: boolean;
+          iw: boolean;
+          vv: boolean;
+          vv_vh: boolean;
+        };
+        orbitDirection: {
+          ascending: boolean;
+          descending: boolean;
+        };
+      };
+      sentinel2: {
+        enabled: boolean;
+        expanded: boolean;
+        l1c: boolean;
+        l2a: boolean;
+        cloudCoverage: number;
+      };
+      sentinel3: {
+        enabled: boolean;
+        expanded: boolean;
+        slstr: boolean;
+        cloudCoverage: number;
+        olci: boolean;
+      };
+      sentinel5P: {
+        enabled: boolean;
+        expanded: boolean;
+        aer_ai: boolean;
+        ch4: boolean;
+        cloud: boolean;
+        co: boolean;
+        hcho: boolean;
+        no2: boolean;
+        o3: boolean;
+        so2: boolean;
+      };
+    };
+    planet: {
+      enabled: boolean;
+      planetScope: {
+        enabled: boolean;
+        expanded: boolean;
+      };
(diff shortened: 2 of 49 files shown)


libs/map/data-access-map/src/lib/data-store/search.model.ts
```diff
@@ -0,0 +1,150 @@
...
+          ew: true,
+          hh: true,
+          hh_hv: true,
+          iw: true,
+          vv: true,
+          vv_vh: true,
+        },
+        orbitDirection: {
+          ascending: true,
+          descending: true,
```
KatarzynaPaszko, line 99: why all those values are true if `enabled` and `extended` are `false`? Won't it cause any issues?

Same to some values below

libs/map/data-access-map/src/lib/data-store/data.model.ts
```diff
@@ -0,0 +1,25 @@
...
+type TInitialData = TSearchData | TSearchDefaultValues | undefined;
+type TUpdateData = TSearchData | undefined;
+
+export type TDataStore<T = TInitialData, U = TUpdateData> = {
+  mode: TMode;
+  schema: TSchema;
+  data: U;
+  state: T;
+  updateData: (data: U) => void;
+  updateState: (data: T) => void;
```
KatarzynaPaszko, line 16:
```suggestion
  updateState: (state: T) => void;
```

KatarzynaPaszko, line 16: Why you add here types `T` and `U` instead of directly set it to `TInitialData` and `TUpdateData` ?

libs/map/data-access-map/src/lib/data-store/search.schema.ts
```diff
@@ -0,0 +1,319 @@
...
+      });
+    }
+  });
+
+const aoiSchema = z.custom<Geometry>((value) => isObject(value));
+
+export const searchSchema = z.object({
+  dataSets: dataSetsSchema,
+  date: dateSchama,
+  aoi: aoiSchema,
```
KatarzynaPaszko, line 316:
```suggestion
  dataSets: dataSetsSchema,
  date: dateSchema,
  aoi: aoiSchema,
```

libs/map/data-access-map/src/lib/data-store/search.schema.ts
KatarzynaPaszko, line 1: This file is huge. Can we divide it somehow?

libs/map/data-access-map/src/lib/data-store/storage/toggle-mode.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+describe('toggleMode', () => {
+  const setup = (defaultState: Partial<TState> = {}, storageName = 'state') => {
+    const memoryStorage = createMemoryStorage();
+    const queryStorage = createMemoryStorage();
+    const toggleMode = createToggleStorage<TData, TData>(memoryStorage, queryStorage)(defaultState, storageName);
+
+    return { toggleMode, memoryStorage, queryStorage };
+  };
+
+  it(`should switch to next mode when there item not exists in Memory Storage`, () => {
```
KatarzynaPaszko, line 49:
```suggestion
  it('should switch to next mode when item does not exists in Memory Storage', () => {
```

libs/map/data-access-map/src/lib/data-store/storage/toggle-mode.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+    const nextSchema = 'action-creator';
+
+    const result = toggleMode(currentState);
+
+    expect(result.mode).toEqual(nextMode);
+    expect(result.schema).toEqual(nextSchema);
+    expect(memoryStorage.getRawItem(currentMode)).toBeNull();
+  });
+
+  it(`should switch to next mode when there item exists in Memory Storage`, () => {
```
KatarzynaPaszko, line 67:
```suggestion
  it('should switch to next mode when item exists in Memory Storage', () => {
```

libs/map/data-access-map/src/lib/data-store/storage/toggle-mode.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+    const { toggleMode, memoryStorage } = setup();
+    const prevMode: TState['mode'] = 'action-creator';
+    const currentState: TState = {
+      mode: 'search',
+      schema: 'search',
+      state: defaultValues,
+    };
+    const prevStorageState: StorageValue<Partial<TState>> = {
+      state: {
+        state: {
```
KatarzynaPaszko, line 77: I would rethink naming  because we have nested  `state` in `state`

libs/map/data-access-map/src/lib/data-store/storage/toggle-mode.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+    memoryStorage.setRawItem(prevMode, prevStorageState);
+    const result = toggleMode(currentState);
+
+    expect(result.mode).toEqual(nextMode);
+    expect(result.schema).toEqual(nextSchema);
+    expect(memoryStorage.getRawItem(prevMode)).toEqual(prevStorageState);
+  });
+
+  describe('while switching to next mode', () => {
+    it(`should clear data when item not exists in Memory Storage`, () => {
```
KatarzynaPaszko, line 104:
```suggestion
    it('should clear data when item does not exists in Memory Storage', () => {
```

KatarzynaPaszko (commented):
There is an issue with build
Screenshot 2024-10-03 at 15 27 25

libs/map/data-access-map/src/lib/data-store/data.model.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type TMode = 'search' | 'action-creator';
+export type TSchema = 'search' | 'action-creator';
+
+type TInitialData = TSearchData | TSearchDefaultValues | undefined;
+type TUpdateData = TSearchData | undefined;
+
+export type TDataStore<T = TInitialData, U = TUpdateData> = {
+  mode: TMode;
+  schema: TSchema;
```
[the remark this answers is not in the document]
mlu-ssoft: the idea is to have multiple schemas. schema is for data validation so in search mode we could have different schemas basing on choosen function

libs/map/data-access-map/src/lib/data-store/search.model.ts
mlu-ssoft: those are default values. by default all sentinels are disabled

libs/map/data-access-map/src/lib/data-store/data.model.ts
mlu-ssoft: because of tests. I had to have option to pass different models there

libs/map/data-access-map/src/lib/data-store/search.schema.ts
mlu-ssoft: I can move aoi and date to separate files, but dataSets which is the biggest part has to be in one file

libs/map/data-access-map/src/lib/data-store/storage/toggle-mode.spec.tsx
mlu-ssoft: first state is internal. second one is property used in the store. I don't think it should be a big problem as it is internal thing

libs/map/data-access-map/src/lib/data-store/storage/toggle-mode.spec.tsx
mlu-ssoft: fixed

libs/map/data-access-map/src/lib/data-store/storage/toggle-mode.spec.tsx
mlu-ssoft: fixed

libs/map/data-access-map/src/lib/data-store/storage/toggle-mode.spec.tsx
mlu-ssoft: fixed

libs/map/data-access-map/src/lib/data-store/data.model.ts
mlu-ssoft: fixed

KatarzynaPaszko (approved):

PR #72, opened 2024-10-08
feat: use aoi, data sets, date range and functions in workflows

mlu-ssoft:
Use data from stores AC:
- AOI
- data sets
- date range
- functions (hardcoded for now, no BE integration yet)

There are still some issues:
- date range isn't cleared

tickets:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-359
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-86
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-90
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-94


libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/value-node.component.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const area = getArea(shape.getExtent());
+  let output;
+  if (area > 10000) {
+    const value = Math.round((area / 1000000) * 100) / 100;
+    output = `${value} km<sup>2</sup>`;
+  } else {
+    const value = Math.round(area * 100) / 100;
+    output = `${value} m<sup>2</sup>`;
+  }
```
KatarzynaPaszko, line 38: Since BE error will be in miles, I think this also should be in miles

libs/map/feature-search-mode-panel/src/lib/use-search-mode.hook.tsx
```diff
@@ -1,50 +1,92 @@
...
+  const values = useMemo(
+    () => ({
+      date,
+      dataSets,
+    }),
+    [date, dataSets]
+  );
+
+  useEffect(() => {
+    // todo move this into data-access layer
```
KatarzynaPaszko, line 46: Whenit will be done? maybe add ticket number

KatarzynaPaszko (commented):
- Please update storybook components, so all Controls that are set would work with current changes

- There is an issue: Given user is in Search mode, perform search and see results, when he switch to Action Creator and then Back to Search Mode, then he sees for instead of Search Results.

@types/i18next/resources.d.ts
```diff
@@ -115,6 +117,9 @@ interface Resources {
             BUTTON_NO: 'No, let me explore on my own';
           };
         };
+        PRESETS: {
+          BUTTON: 'Load preset';
+        };
```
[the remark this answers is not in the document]
mlu-ssoft: I didn't. I generated this file, I think that it wasn't updated

libs/map/feature-search-mode-panel/src/lib/use-search-mode.hook.tsx
mlu-ssoft: we don't have ticket number. I don't know when and in which ticket it will be done. it is an information that when we will start working on this component we should think about moving it into data-access layer. Well probably when we will integrate new data into search view but i'm not sure

libs/map/data-access-map/src/lib/aoi-store/geometry.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    case 'circle': {
+      return new Circle(coordinate.center, coordinate.radius);
+    }
+
+    default: {
+      return undefined;
+    }
+  }
+};
```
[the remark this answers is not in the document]
mlu-ssoft: fixed

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/value-node.component.tsx
mlu-ssoft: fixed

libs/map/feature-search-mode-panel/src/lib/use-search-mode.hook.tsx
mlu-ssoft: fixed

KatarzynaPaszko (approved):

PR #74, opened 2024-10-11
UKRIW-398 - displaying Presets functions

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-398

This pull request introduces several changes to improve the handling of presets, enhance the HTTP client configuration, and update the environment configuration. The most important changes include adding a new internal API URL, implementing a new preset schema and query, and refactoring various components to support these updates.

### Environment Configuration Updates:
* Added `VITE_INTERNAL_API_URL` to `.env`.
* Updated `interface IEnvConfig` to include `internalApiUrl`.
* Modified `getEnvConfig` to handle `VITE_INTERNAL_API_URL`.

### HTTP Client Configuration:
* Updated `initHttpClient` function to support multiple base URLs by adding an `id` parameter.

### Preset Handling:
* Introduced `presetsSchema` and `TPresets` type.
* Implemented `useGetPresets` query and used it in `Presets` component.
* Added error handling and Loading spinner

### Component Refactoring:
* Updated`Footer` component to handle conditional rendering based on the active tab.
* Updated `Content` component to include the new `Footer`.

### Other Changes:
* Updated `img-src` directive in `apps/eodh-fe/index.html` to include `data:`.


libs/shared/utils/react-query/src/lib/axios.ts
```diff
@@ -9,34 +9,33 @@ type TRequestConfig = {
   params?: IHttpRequest['params'];
 };
 
-export const getHttpClient = () => {
+const instances: { [key: string]: AxiosInstance } = {};
+
+export const getHttpClient = (id = 'baseUrl') => {
```
mlu-ssoft, line 14: this should be generic, it shouldn't know anything about different implementations (library should be configurable, but it shouldn't know nothing about specific implementations). it means that 'baseUrl' default parameter value shouldn't be here, it should came from configuration.

libs/shared/utils/react-query/src/lib/axios.ts
```diff
@@ -46,6 +45,10 @@ export const initHttpClient = (config: IHttpClientConfig, interceptors: IHttpInt
   applyInterceptors(instance, interceptors);
 
   instance.interceptors.response.use((response) => response.data);
+
+  instances[id] = instance;
```
mlu-ssoft, line 49: what if instance already exists?

scripts/create-config-file.js
```diff
@@ -8,6 +8,7 @@ function getVariableValue(variable, defaultValue = '') {
 const config = {
   baseUrl: getVariableValue(process.env.VITE_BASE_URL),
   apiUrl: getVariableValue(process.env.VITE_API_URL),
+  internalApiUrl: getVariableValue(process.env.VITE_INTERNAL_API_URL),
```
mlu-ssoft, line 11: move this to `httpClient` config

libs/shared/utils/react-query/src/lib/model.ts
```diff
@@ -18,5 +18,6 @@ export interface IHttpInterceptor {
 }
 
 export interface IHttpClientConfig {
+  id: string;
```
mlu-ssoft, line 21: it should be array with available urls, eg `proxyConfig: { [key: string]: string }`

apps/eodh-fe/index.html
```diff
@@ -13,7 +13,7 @@
         worker-src  blob: %VITE_BASE_URL%/e1580240-4509-447d-b8af-40ec4206f933;
         font-src    'self';
         script-src  'self';
-        img-src     'self' https://tile.openstreetmap.org https://sentinel-cogs.s3.us-west-2.amazonaws.com;
+        img-src     'self' https://tile.openstreetmap.org https://sentinel-cogs.s3.us-west-2.amazonaws.com data:;
```
mlu-ssoft, line 16: data shouldn't be allowed from one domain only?

apps/eodh-fe/.env.example
```diff
@@ -1,5 +1,6 @@
 VITE_BASE_URL=http://example.com
 VITE_API_URL=http://example.com
+VITE_INTERNAL_API_URL=http://example.com
```
mlu-ssoft, line 3: I would use different name, maybe?:
- STAC or TPZ/TELESPAZIO - for telespazio API
- EOPRO - for our API

mlu-ssoft (changes requested):

apps/eodh-fe/index.html
KatarzynaPaszko: this was added so we can display images base64

apps/eodh-fe/.env.example
mlu-ssoft: not fixed, change also names here

apps/eodh-fe/index.html
mlu-ssoft: yep, I know, but shouldn't be allow using data from only one domain? I don't know if it is possible, but if it is it will be better from security point of view

apps/eodh-fe/src/env.config.ts
```diff
@@ -54,6 +57,15 @@ const getValue = <T extends string | string[] | undefined[] | boolean>(
   return defaultValue;
 };
 
+const eodhProUrl = () => {
```
mlu-ssoft, line 60: maybe `getEodhProUrl`? this is function, not constant

apps/eodh-fe/src/env.config.ts
```diff
@@ -54,6 +57,15 @@ const getValue = <T extends string | string[] | undefined[] | boolean>(
...
+  const apiVersion = '/v1.0';
+  const importedUrl = getValue<string>(import.meta.env.VITE_EODH_PRO_API_URL, config?.authorization.clientId, '');
+  if (importedUrl) {
```
mlu-ssoft, line 63: add one line before for better readibility

apps/eodh-fe/src/env.config.ts
```diff
@@ -54,6 +57,15 @@ const getValue = <T extends string | string[] | undefined[] | boolean>(
...
+    return importedUrl.replace(/\/+$/, '') + apiVersion;
+  }
+  return '';
```
mlu-ssoft, line 66: add one line before for better readibility

apps/storybook-host/.storybook/config/env.config.ts
```diff
@@ -54,6 +57,15 @@ const getValue = <T extends string | string[] | undefined[] | boolean>(
   return defaultValue;
 };
 
+const eodhProUrl = () => {
+  const apiVersion = '/v1.0';
+  const importedUrl = getValue<string>(import.meta.env.VITE_EODH_PRO_API_URL, config?.authorization.clientId, '');
+  if (importedUrl) {
```
mlu-ssoft, line 63: add one line before for better readibility

apps/storybook-host/.storybook/config/env.config.ts
```diff
@@ -54,6 +57,15 @@ const getValue = <T extends string | string[] | undefined[] | boolean>(
...
+    return importedUrl.replace(/\/+$/, '') + apiVersion;
+  }
+  return '';
```
mlu-ssoft, line 66: add one line before for better readibility

libs/map/data-access-map/src/lib/presets/presets.model.ts
```diff
@@ -0,0 +1,40 @@
...
+  default: z.union([z.string(), z.null()]),
+  options: z.union([z.array(z.string()), z.null()]),
+});
+
+const outputSchema = z.object({
+  type: z.string(),
+  description: z.string(),
+});
+
+const functionSchema = z.object({
```
mlu-ssoft, line 16: endpoint with this schema returns both presets and functions. we should have one react query hook which will return cached data and two additional hooks for presets and functions (those should filter by preset field). I think we should have different schema (internal) for presets and functions. In preset we don't need to have inputs field, only name, description etc will be important

libs/map/data-access-map/src/lib/presets/response-formatter.ts
```diff
@@ -0,0 +1,11 @@
+import { TPresets } from './presets.model';
+
+export const formatPresetsResponse = (response: TPresets): TPresets => {
+  response.functions.forEach((func) => {
```
mlu-ssoft, line 4: you can do this using zod

libs/map/data-access-map/src/lib/presets/presets.model.ts
```diff
@@ -0,0 +1,40 @@
...
+    date_start: inputSchema.optional(),
+    date_end: inputSchema.optional(),
+    aoi: inputSchema.optional(),
+    bbox: inputSchema.optional(),
+    index: inputSchema.optional(),
+  }),
+  outputs: z.object({
+    collection: outputSchema,
+  }),
+});
```
mlu-ssoft, line 33: I would add map this type into internal one, remove `thumbnail_b64` and create `image`/`imageUrl` property instead

libs/map/feature-action-creator-panel/src/lib/content/presets/preset.component.tsx
```diff
@@ -1,13 +1,14 @@
 import { Button, Icon, Text } from '@ukri/shared/design-system';
+import clsx from 'clsx';
 import { useCallback, useState } from 'react';
 
 import { presetStyles } from './preset.styles';
 
 interface IImageProps {
-  imageUrl: string;
+  base64String: string;
```
mlu-ssoft, line 8: is it still `base64String`? You added data attribute in data-access, I think `imageUrl` is right now better name

libs/map/feature-action-creator-panel/src/lib/content/presets/preset.component.tsx
```diff
@@ -24,26 +25,27 @@ const Image = ({ imageUrl }: IImageProps) => {
...
   title: string;
-  description: string;
+  description?: string;
   onLoadPresetClick: () => void;
+  className?: string;
 }
 
-export const Preset = ({ imageUrl, title, description, onLoadPresetClick }: IResultItemProps) => {
+export const Preset = ({ base64String, title, description, onLoadPresetClick, className }: IResultItemProps) => {
```
mlu-ssoft, line 41: same here, I think `imageUrl` is better than `base64String`

libs/shared/utils/react-query/src/lib/axios.ts
```diff
@@ -15,37 +16,32 @@ export const getHttpClient = () => {
...
     return;
   }
 
-  applyInterceptors(instance, interceptors);
+  instance = Axios.create({ timeout: 60000 });
+  applyInterceptors(instance, [new ProxyInterceptor(config), ...interceptors]);
 
   instance.interceptors.response.use((response) => response.data);
+
+  return instance;
```
mlu-ssoft, line 44: this fn should be void, remove

libs/shared/utils/react-query/src/lib/proxy.interceptor.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  public handleError = (error: IHttpErrorResponse): Promise<IHttpErrorResponse> => {
+    return Promise.reject(error);
+  };
+
+  private handleRequest = (request: IHttpRequest, proxyConfig?: TProxyConfig): IHttpRequest => {
+    if (!proxyConfig || Object.keys(proxyConfig).length === 0) {
+      return request;
+    }
+    const arrayOfUrls = Object.entries(proxyConfig)
```
mlu-ssoft, line 18: add one empty line before for better readibility

libs/shared/utils/react-query/src/lib/proxy.interceptor.ts
```diff
@@ -0,0 +1,26 @@
...
+  public async intercept(request: IHttpRequest): Promise<IHttpRequest> {
+    return this.handleRequest(request, this.config.proxyConfig);
+  }
+
+  public handleError = (error: IHttpErrorResponse): Promise<IHttpErrorResponse> => {
+    return Promise.reject(error);
+  };
+
+  private handleRequest = (request: IHttpRequest, proxyConfig?: TProxyConfig): IHttpRequest => {
+    if (!proxyConfig || Object.keys(proxyConfig).length === 0) {
```
mlu-ssoft, line 15: it can be simplified: `!Object.keys(proxyConfig).length`

libs/shared/utils/react-query/src/lib/proxy.interceptor.ts
```diff
@@ -0,0 +1,26 @@
...
+      return request;
+    }
+    const arrayOfUrls = Object.entries(proxyConfig)
+      .map(([key]) => {
+        return request.url?.startsWith(key) ? request.url?.replace(key, proxyConfig[key]) : undefined;
+      })
+      .filter((url) => !!url);
+
+    return { ...request, url: arrayOfUrls.length > 0 ? arrayOfUrls.pop() : request.url };
```
mlu-ssoft, line 24: it can be simplified: `arrayOfUrls.length ? arrayOfUrls.pop() : request.url`

libs/shared/utils/react-query/src/index.ts
```diff
@@ -2,3 +2,4 @@ export { createReactQueryWrapper, queryClient } from './lib/react-query';
 export type { TExtractFnReturnType, TMutationConfig, TQueryConfig } from './lib/react-query';
 export { withQueryClient } from './lib/with-query-client';
 export { initHttpClient, getHttpClient } from './lib/axios';
+export type { IHttpClientConfig, IHttpInterceptor, IHttpRequest } from './lib/model';
```
mlu-ssoft, line 5: is it used somewhere, do we need to export this? I couldn't find imports for those types

libs/shared/utils/react-query/src/lib/proxy.interceptor.ts
mlu-ssoft: it can be simplified:
`Object.values(proxyConfig).map((key) => request.url?.startsWith(key) ? request.url?.replace(key, proxyConfig[key]) : undefined).filter((url) => !!url);`

mlu-ssoft: this still don't check if there are trailing slashes:
- I can add multiple trailing slashes in the configuration, eg: `https://test.eodatahub.org.uk//////`
- I'm not forced to use types. I can forgot about it and pass url like that: `/////example-url`
library should prevent such situations

apps/eodh-fe/src/env.config.ts
mlu-ssoft: what if `importedUrl` will have trailing slash, eg `https://test.eodatahub.org.uk////`? if you are combining two strings you should make sure that there won't be any unnecessary slashes

mlu-ssoft (changes requested):

libs/shared/utils/react-query/src/lib/proxy.interceptor.ts
KatarzynaPaszko: > it can be simplified: `Object.values(proxyConfig).map((key) => request.url?.startsWith(key) ? request.url?.replace(key, proxyConfig[key]) : undefined).filter((url) => !!url);`

It wont work with `Object.values`, thats why I have updated it with `Object.entries`

apps/eodh-fe/index.html
KatarzynaPaszko: I tried to find other way, but Im keep on getting CSP error. Do you know some other way to overcome this issue?

mlu-ssoft (approved):

PR #76, opened 2024-10-18
UKRIW-410 - I want to be able to use history tab

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-410

This pull request introduces a new history feature to the application, including changes to the data access layer, UI components, and translations. The changes enable users to view, sort, and filter workflow history and load more results if there are more than 1 page.


libs/map/data-access-map/src/lib/history/history.model.ts
```diff
@@ -0,0 +1,45 @@
+import { z } from 'zod';
+
+const statusEnum = z.enum(['submitted', 'running', 'cancel-request', 'successful', 'failed', 'cancelled']);
+
+const submissionSchema = z.object({
+  submission_id: z.string().uuid(),
+  status: statusEnum.transform((status) => {
+    if (['failed', 'cancel-request', 'cancelled'].includes(status)) {
+      return 'FAILED';
```
mlu-ssoft, line 9: 1. do we need to use native BE statuses somewhere?
2. how do we mark item as ready on websockets?
3. how and where mapping on websockets will be done?

libs/map/data-access-map/src/lib/history/history.model.ts
```diff
@@ -0,0 +1,45 @@
...
+      return 'PROCESSING';
+    } else if (status === 'successful') {
+      return 'READY';
+    }
+  }),
+  function_identifier: z.string(),
+  submitted_at: z
+    .string()
+    .datetime()
+    .transform((dateString) => {
```
mlu-ssoft, line 20: I wouldn't map it here. It may be like that in other places we will need full date. In theory this endpoint may be used by multiple views. This mapping (into date and hour) looks like very specific to one view. I would do mapping in this view then (instead of doing it globally)

libs/map/data-access-map/src/lib/history/history.model.ts
```diff
@@ -0,0 +1,45 @@
...
+      const date = new Date(dateString);
+      return {
+        date: date.toISOString().split('T')[0],
+        hour: `${date.getUTCHours().toString().padStart(2, '0')}:${date.getUTCMinutes().toString().padStart(2, '0')}`,
+      };
+    }),
+  finished_at: z.union([z.string().datetime(), z.null(), z.undefined()]),
```
mlu-ssoft, line 27: we are using camelCase, I would mapp `finished_at` into `finishedAt`. same with other fields

libs/map/data-access-map/src/lib/history/history.query.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { historySchema, THistory } from './history.model';
+
+export interface IHistoryParams {
+  orderBy?: string;
+  orderDirection?: 'asc' | 'desc';
+  page?: number;
+  perPage?: number;
+}
+
+const history = async ({
```
mlu-ssoft, line 14: maybe `getHistoryResults`?

libs/map/feature-action-creator-panel/src/lib/content/history/history-tile/history-tile.component.tsx
```diff
@@ -0,0 +1,99 @@
...
+}: IHistoryTileProps) => {
+  const { t } = useTranslation();
+
+  return (
+    <div className={clsx(historyTileStyles.container(selected), className)}>
+      <div className={historyTileStyles.section}>
+        <div className={historyTileStyles.textContainer}>
+          <Text content={function_identifier} fontSize='medium' fontWeight='semibold' />
+          <Text
+            content={`ID: ${truncateString(workflowId, 25)}`}
```
mlu-ssoft, line 63: add it into translations. maybe in different language `ID` will be something different

libs/map/feature-action-creator-panel/src/lib/content/history/history-tile/history-tile.component.tsx
```diff
@@ -0,0 +1,99 @@
...
+        {status && <Tag status={status} />}
+        {status !== 'FAILED' && (
+          <Button
+            text={
+              selected
+                ? 'MAP.ACTION_CREATOR_PANEL.HISTORY.HIDE_RESULTS'
+                : 'MAP.ACTION_CREATOR_PANEL.HISTORY.VIEW_RESULTS'
+            }
+            size='medium'
+            onClick={selected ? onHideResult : onViewResult}
```
mlu-ssoft, line 92: maybe create two buttons? SelectedButton and ViewResultsButton? There is logic repeated twice (in text and in onClick)

libs/map/feature-action-creator-panel/src/lib/content/history/sort-filter/sort-filter.component.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+type TSortOrder = 'newest' | 'oldest';
+
+interface ISortFilterProps {
+  onSortChange: (order: TSortOrder) => void;
+  sortKey: 'default' | 'newest' | 'oldest';
+  className?: string;
+}
+
+export const SortFilter = React.memo(({ onSortChange, sortKey, className }: ISortFilterProps) => {
```
mlu-ssoft, line 16: you can import `memo` from React lib directly

libs/map/feature-action-creator-panel/src/lib/content/history/sort-filter/sort-filter.component.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const dropdownRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    const handleClickOutside = (event: MouseEvent) => {
+      if (dropdownRef.current && !dropdownRef.current.contains(event.target as Node)) {
+        setIsOpen(false);
+      }
+    };
+
+    document.addEventListener('mousedown', handleClickOutside);
```
mlu-ssoft, line 29: don't we have dropdown component with clicking outside logic already? I prefer do not use native browser features in multiple places if not needed. eg if at some point we would like to create mobile app then it will be hard to find all those places

libs/map/feature-action-creator-panel/src/lib/content/history/sort-filter/sort-filter.component.tsx
```diff
@@ -0,0 +1,76 @@
...
+    return () => {
+      document.removeEventListener('mousedown', handleClickOutside);
+    };
+  }, []);
+
+  const toggleDropdown = () => setIsOpen(!isOpen);
+
+  const handleSortSelection = (order: TSortOrder) => {
```
mlu-ssoft, line 37: add useCallback

libs/map/feature-action-creator-panel/src/lib/content/history/sort-filter/sort-filter.component.tsx
```diff
@@ -0,0 +1,76 @@
...
+      }
+    };
+
+    document.addEventListener('mousedown', handleClickOutside);
+    return () => {
+      document.removeEventListener('mousedown', handleClickOutside);
+    };
+  }, []);
+
+  const toggleDropdown = () => setIsOpen(!isOpen);
```
mlu-ssoft, line 35: add useCallback

libs/map/feature-action-creator-panel/src/lib/content/history/use-history-data.hook.ts
```diff
@@ -0,0 +1,69 @@
...
+  const params: IHistoryParams = {
+    orderBy: 'submitted_at',
+    orderDirection: sortOrder,
+    page: page,
+    perPage: 25,
+  };
+
+  const { data, error, isLoading, isFetching, refetch } = useGetHistory(params);
+
+  const loadMore = () => {
```
mlu-ssoft, line 41: add useCallback

libs/map/feature-action-creator-panel/src/lib/content/history/use-history-data.hook.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (data && data.results.length > 0) {
+      setPage((prevPage) => prevPage + 1);
+    }
+  };
+
+  useEffect(() => {
+    if (data && !isLoading && !isFetching) {
```
mlu-ssoft, line 48: I think you could use `isPending` which is the same as `isLoading || isFetching`

libs/map/feature-action-creator-panel/src/lib/content/history/use-history-data.hook.ts
```diff
@@ -0,0 +1,69 @@
...
+      const newResults = data.results.filter(
```
mlu-ssoft, line 49: this state right now is locall. how do we mark item as ready when we will have websockets?

mlu-ssoft (changes requested):

libs/map/data-access-map/src/lib/history/history.model.ts
KatarzynaPaszko: 1. - we just need to know if it is failed or no to display red tag, thats why I map all 3 to one here
2. , 3.  - it was agreed that it will be done in this ticket  https://ukri-spyrosoft.atlassian.net/browse/UKRIW-117

libs/map/feature-action-creator-panel/src/lib/content/history/sort-filter/sort-filter.component.tsx
KatarzynaPaszko: This is totally different component with different logic and styling than dropdown that we have in DesignSystem

libs/map/feature-action-creator-panel/src/lib/content/history/use-history-data.hook.ts
KatarzynaPaszko: This will need to be done in https://ukri-spyrosoft.atlassian.net/browse/UKRIW-117

libs/map/data-access-map/src/lib/history/history.query.tsx
KatarzynaPaszko: With presets and search we just used one word. Do you want to change convention everywhere?
Screenshot 2024-10-21 at 11 55 52
Screenshot 2024-10-21 at 11 56 09

libs/map/feature-action-creator-panel/src/lib/content/history/use-history-data.hook.ts
KatarzynaPaszko: `IsPending` was introduced to replace `isLoading`, but not `isFetching`: https://tanstack.com/query/v5/docs/framework/react/guides/migrating-to-v5

I will replace `isLoading` with `isPending`

libs/map/feature-action-creator-panel/src/lib/content/history/history-tile/history-tile.component.tsx
mlu-ssoft: yep, and it is logic, so not only text but also onClick behaviour changed.
selected condition is repeated twice, so you can add:
- two buttons and display only one of them at a time
- create separate component for button only

It will be more visible then that basing on selection we have different buttons

libs/map/feature-action-creator-panel/src/lib/content/history/sort-filter/sort-filter.component.tsx
mlu-ssoft: 1. why it is different? why do we have both?
2. can we create different versions of the same component (in design system)?

libs/map/feature-action-creator-panel/src/lib/content/history/use-history-data.hook.ts
mlu-ssoft: I'm not asking about the task. You know that in the future we will implement websockets so implementation should support it from start. If you don't do this then we will have to rewrite implementation. You should think about the whole implementation (websockets, possible future changes, will we need BE statuses in the future, etc). Then create solution which support all of it, eg while implementing websockets we should only update status somewhere. We shouldn't be forced to rewrite current implementetion every time requirements change a little bit

libs/map/feature-action-creator-panel/src/lib/content/presets/presets.component.tsx
```diff
@@ -19,15 +19,15 @@ const ErrorMessage = ({ refetch }: IErrorMessageProps) => (
...
+  const { data, error, isPending, isFetching, refetch } = useGetPresets();
 
   const handleLoadPreset = (preset: unknown) => {
     // eslint-disable-next-line no-console
     console.log('Load preset', preset);
     return;
   };
 
-  if (isLoading || isFetching) {
+  if (isPending || isFetching) {
```
mlu-ssoft, line 30: if you will use `isLoading` only then you won't see spinner twice (request sill will be made, but data from cache will be displayed to the user and replaced under the hood when call is finished)

mlu-ssoft (changes requested):

libs/map/feature-action-creator-panel/src/lib/content/history/sort-filter/sort-filter.component.tsx
KatarzynaPaszko: As discussed on call, it is different visually and with logic. 

I have extracted document.addEventListener mouse down logic to separate hook `useOutsideClick`

libs/map/data-access-map/src/lib/history/history.model.ts
mlu-ssoft: I don't know, for now nowhere, but in the future we may use it somewhere else. We shouldn't be forced to change implementation every time we will be adding new feature. Data access layer can be used by multiple views and it should be treatten like that (one to many relationship). For now it is just one view, but in the future we might have features which will use this data too. It is cached data so you should just take it and do something in the view (if needed). The same applies to statuses: will we need BE statuses in different places in the future? If we will have different view (something similar to history tab), what is the chance that we will have exact same statuses as they are currently are in history tab? Where is better to do mapping (in data-access layer or in the component itself)? Those are questions that should be asked/answered. So you should cover not only this exaclt implelmentation but create code which will allow us to easilly extend/add new functionalities without needing of refactor/change current ones

libs/map/data-access-map/src/lib/history/history.query.tsx
mlu-ssoft: yep, you can change it also there

libs/shared/utils/date/src/lib/date.utils.ts
```diff
@@ -37,10 +39,23 @@ export const formatDate = (date: TDateTimeString, format: 'YYYY-MM-DD' = 'YYYY-M
   switch (format) {
     case 'YYYY-MM-DD': {
       dateString = createIsoStringDate(date).split('T')[0];
+      break;
+    }
+    case 'DD-MM-YY': {
```
mlu-ssoft, line 44: shouldn't it be DD/MM/YYYY? I added it in another PR

mlu-ssoft (approved):

libs/shared/utils/date/src/lib/date.utils.ts
KatarzynaPaszko: I added it as per design:
Screenshot 2024-10-22 at 09 38 01

PR #77, opened 2024-10-21
Feature/ukriw 489 fix drawing tools enabled after switching to Action Creator mode 

mlu-ssoft:
- fixed drawing tools enabled by default after switching to AC mode
- adjusted colors to met designs

tickets:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-489
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-492
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-500
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-506

### DIFF
--- a/libs/map/feature-search-mode-panel/src/lib/use-search-mode.hook.tsx
+++ b/libs/map/feature-search-mode-panel/src/lib/use-search-mode.hook.tsx
@@ -87,7 +87,11 @@ export const useSearchMode = () => {
       if (searchParams) {
         changeView('results');
       } else {
-        changeView('search');
+        if (mode === 'action-creator') {
+          setView('search');
+        } else {
+          changeView('search');
+        }
       }
       setCurrentMode(mode);
     }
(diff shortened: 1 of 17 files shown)


libs/map/feature-search-mode-panel/src/lib/use-search-mode.hook.tsx
```diff
@@ -87,7 +87,11 @@ export const useSearchMode = () => {
       if (searchParams) {
         changeView('results');
       } else {
-        changeView('search');
+        if (mode === 'action-creator') {
+          setView('search');
+        } else {
+          changeView('search');
+        }
```
KatarzynaPaszko, line 94: What's the difference between setView and changeView ?

mlu-ssoft: setView is a useState fn, it is changing local view. ChangeView is doing more things, it is calling setView fn and also updating AoiStore state. I changed fn name into setCurrentView, I think right now it should be more explanatory

PR #82, opened 2024-10-23
UKRIW-410 -  updated displayed date and time to be local date/time

KatarzynaPaszko:
updated displayed date and time to be local date/time not UTC date/time

### DIFF
--- a/libs/map/feature-action-creator-panel/src/lib/content/history/history-tile/history-tile.component.tsx
+++ b/libs/map/feature-action-creator-panel/src/lib/content/history/history-tile/history-tile.component.tsx
@@ -75,7 +75,6 @@ export const HistoryTile = ({
   className,
 }: IHistoryTileProps) => {
   const { t } = useTranslation();
-  const submittedAt = new Date(submittedAtDate);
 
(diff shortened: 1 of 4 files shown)


libs/map/feature-action-creator-panel/src/lib/content/history/history-tile/history-tile.component.tsx
```diff
@@ -77,8 +77,8 @@ export const HistoryTile = ({
   const { t } = useTranslation();
   const submittedAt = new Date(submittedAtDate);
 
-  const submittedHour = `${submittedAt.getUTCHours().toString().padStart(2, '0')}:${submittedAt
-    .getUTCMinutes()
+  const submittedHour = `${submittedAt.getHours().toString().padStart(2, '0')}:${submittedAt
```
mlu-ssoft, line 80: move this into `shared/utils/date`. then we will have one place for managing dates. you can create `time.ts` file (or similar) and put it there. Do similar as with `date.utils` - create separate format for that which should be used in the application

mlu-ssoft (changes requested):

mlu-ssoft (approved):

PR #85, opened 2024-10-25
[UKRIW-367] load preset

mlu-ssoft:
Load Preset into Workflow

Ticket:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-471

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -4,6 +4,10 @@ on:
       - 'develop'
       - 'main'
 
+permissions:
+  id-token: write
+  contents: read
+
 jobs:
   build:
     name: 'Build'
@@ -11,3 +15,20 @@ jobs:
   buildStorybook:
     name: 'Build Storybook'
     uses: ./.github/workflows/storybook-build.yml
+  deploy:
+    name: 'Deploy'
+    uses: ./.github/workflows/deploy.yml
+    needs: [build]
+    with:
+      canDeployOnDev: true
+      canDeployOnQA: true
+      canDeployOnStaging: false
+      canDeployOnProduction: false
+    secrets: inherit
+  deployStorybook:
+    name: 'Deploy Storybook'
+    uses: ./.github/workflows/storybook-deploy.yml
+    needs: [buildStorybook]
+    with:
+      canDeploy: true
+    secrets: inherit
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -4,10 +4,6 @@ on:
       - 'develop'
       - 'main'
 
-permissions:
-  id-token: write
-  contents: read
-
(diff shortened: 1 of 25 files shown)


.github/workflows/pull-request.yml
KatarzynaPaszko, line 1: How those changes are connected to loading Presets?

libs/map/data-access-map/src/lib/query/presets.model.ts
```diff
@@ -21,15 +28,15 @@ const numberSchema = z.object({
...
 const inputSchema = z.union([booleanSchema, stringSchema, numberSchema]);
 
-const functionSchema = z
+const presetSchema = z
   .object({
     name: z.string(),
     preset: z.boolean(),
     identifier: z.string(),
     description: z.string().optional(),
     thumbnail_b64: z.string(),
```
KatarzynaPaszko, line 37: thumbnail also should be optional

libs/map/data-access-map/src/lib/query/presets.model.ts
```diff
@@ -21,15 +28,15 @@ const numberSchema = z.object({
...
     inputs: z.object({
-      stac_collection: inputSchema.optional(),
+      stac_collection: collectionSchema.optional(),
       calibrate: inputSchema.optional(),
       index: inputSchema.optional(),
       limit: inputSchema.optional(),
```
KatarzynaPaszko, line 42: calibrate, index and limit should all have `options`, so all of it should be of type `collectionSchema`

libs/map/data-access-map/src/lib/store/data-sets/data-sets.store.tsx
KatarzynaPaszko, line 1: Can we not hardcode everything here and make it more dynamic ?

mlu-ssoft: we don't have time to do it right now. It will take at least few days

libs/map/data-access-map/src/lib/query/presets.model.ts
mlu-ssoft: the difference is that in `inputSchema` `default` can be optional where in `stat_collection` not. I changed it (removed collectionSchema)

KatarzynaPaszko: so please add `options` to `inputSchema` as well.

mlu-ssoft: but there are options in `inputSchema`. `inputSchema` can be:
- stringSchema (it has options)
- booleanSchema (it doesn't have options)
- numberSchema (it doesn't have options)

KatarzynaPaszko (approved):

PR #87, opened 2024-10-28
Feature/ukirw 442 add error messages for running workflow

mlu-ssoft:
Add display errors for running Workflow.

ticket:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-442


libs/map/data-access-map/src/lib/mutation/workflow.error.tsx
```diff
@@ -0,0 +1,187 @@
...
+      return 'MAP.ACTION_CREATOR_PANEL.WORKFLOW.NODE.DATA_SET.AUXILIARY.CORINE_LAND_COVER';
+    }
+
+    case 'clms-water-bodies': {
+      return 'MAP.ACTION_CREATOR_PANEL.WORKFLOW.NODE.DATA_SET.AUXILIARY.WATER_BODIES';
+    }
+  }
+
+  return null;
+};
```
KatarzynaPaszko, line 94:
```suggestion
const BASE_KEY = 'MAP.ACTION_CREATOR_PANEL.WORKFLOW.NODE.DATA_SET';

const collectionTranslationMap: Record<TCollectionName, string> = {
  'sentinel-1': `${BASE_KEY}.SENTINEL_1`,
  'sentinel-2-l1c': `${BASE_KEY}.SENTINEL_2`,
  'sentinel-2-l2a': `${BASE_KEY}.SENTINEL_2`,
  'sentinel-3': `${BASE_KEY}.SENTINEL_3`,
  'sentinel-5p': `${BASE_KEY}.SENTINEL_5P`,
  'esacci-globallc': `${BASE_KEY}.AUXILIARY.GLOBAL_LAND_COVER`,
  'clms-corinelc': `${BASE_KEY}.AUXILIARY.CORINE_LAND_COVER`,
  'clms-water-bodies': `${BASE_KEY}.AUXILIARY.WATER_BODIES`
};

const getCollectionTranslationKey = (collection: TCollectionName): string | null => {
  return collectionTranslationMap[collection] || null;
};
```

KatarzynaPaszko: Do similar with below cases as well

libs/shared/design-system/src/lib/snackbar/snackbar-provider.component.tsx
```diff
@@ -0,0 +1,25 @@
...
+      }}
+      anchorOrigin={{
+        vertical: 'bottom',
+        horizontal: 'center',
+      }}
+      action={(snackbarKey) => <SnackbarCloseButton snackbarKey={snackbarKey} />}
+    >
+      {children}
+    </NotistackProvider>
+  );
```
KatarzynaPaszko, line 24: you can use here our Notification component, so then you do not have to overwrite all styling in _libs/shared/design-system/src/lib/snackbar/snackbar.css_

KatarzynaPaszko (approved):

apps/eodh-fe/public/assets/i18n/en.json
```diff
@@ -157,6 +121,12 @@
         }
       },
       "WORKFLOW": {
+        "ERROR": {
+          "AOI_TOO_BIG": "Area exceeds {{maxSize}} square miles.",
```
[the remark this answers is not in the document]
mlu-ssoft: I would leave it as it is for now. If we will add support for it then we will add second key for translations. Right now there are two places which are displaying units. To do this properly we would have to add setting in the app to check which metric is chooses. Its not to big deal but right now we have a lot of work anyway, we can easilly add this in the next sprint

libs/shared/design-system/src/lib/snackbar/snackbar-provider.component.tsx
mlu-ssoft: will it work with Notistack API? Notistack can display multiple actions.

PR #88, opened 2024-10-30
UKRIW-120 - display ac workflow results

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-120


libs/map/feature-action-creator-panel/src/lib/content/history/history-tile/history-tile.hook.tsx
```diff
@@ -0,0 +1,74 @@
...
+      }
+      if (data) {
+        setAcResultsData(data);
+        setAcResultStatus('success');
+      }
+    }
+  }, [data, error, isFetching, isPending, jobId, setAcResultStatus, setAcResultsData]);
+
+  const handleSeeResults = useCallback(() => {
+    changeAcView('acResultsView');
```
mlu-ssoft, line 38: whats the difference between `acResultsView` and `resultsView`? Can't we have just one `resultsView`?

libs/map/feature-action-creator-panel/src/lib/content/history/history-tile/history-tile.hook.tsx
```diff
@@ -0,0 +1,74 @@
...
+  workflowId: string,
+  submittedAtDate: string,
+  onViewResult: (submissionId: string) => void,
+  onHideResult: () => void
+) => {
+  const [jobId, setJobId] = useState<string | null>(null);
+  const { t } = useTranslation();
+  const { userWorkspace } = useAuth();
+  const { isPending, isFetching, data, error } = useGetWorkflowResult({ jobId, userWorkspace });
+  const { setAcResultsData, setAcResultStatus, changeAcView } = useMode();
```
mlu-ssoft, line 17: why you are setting status separatelly? You have it already in the query. Check how querying for Search is done - it is enough to just make a call one again (with parameter).

libs/map/data-access-map/src/lib/query/workflow-result.query.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import { paths } from '../api';
+import { collectionSchema, TCollection } from './workflow-result.model';
+
+export interface IWorkflowResultsParams {
+  jobId: string | null;
+  userWorkspace?: string;
+}
+
+const getWorkflowResults = async ({ jobId, userWorkspace }: IWorkflowResultsParams): Promise<TCollection> => {
```
mlu-ssoft, line 12: I would reuse `QueryBuilder` from `data-access-stac-catalog`. This API have the same contract (you can filter it by complex queries which may be done in the future). We shouldn't have multiple instances which will deal with the same API contract. I would excent `QueryBuilder` to support multiple endpoints and data sets (eg those for land cover changes)

libs/map/data-access-map/src/lib/store/mode.model.ts
```diff
@@ -1,7 +1,15 @@
+import type { TCollection } from '../query/workflow-result.model';
 export type TMode = 'search' | 'action-creator';
+export type TACView = 'acTreeView' | 'acResultsView';
 
 export interface IModeStore {
   mode: TMode;
   changeMode: (mode: TMode) => void;
   toggleMode: () => void;
+  acView: TACView;
+  changeAcView: (view: TACView) => void;
```
mlu-ssoft, line 10: there is store for results view already: `results.store` in `libs/map/data-access-map/src/lib/store/results/results.store.tsx`, you could use it. If you will send new parameters it should trigger the query. I would try to use it first

libs/map/feature-search-mode-panel/src/lib/search-mode-panel.component.tsx
```diff
@@ -28,7 +40,24 @@ export const SearchModePanel = () => {
       );
     }
 
+    case 'acResultsView': {
```
mlu-ssoft, line 43: why to have separate action for that? You could reuse `results` case but with different parameters:
- add support for multiple endpoints in QueryBuilder
- use different parameters so QueryBuilder will know which endpoint it should use. It will automatically send request and refresh the view

libs/map/ui-results-view/src/lib/results-list/result-item/result-item.tsx
```diff
@@ -83,8 +81,8 @@ export const ResultItem = ({
   onRemoveFromCompare,
 }: IResultItemProps) => {
   const [isAddedForComparison, setIsAddedForComparison] = useState(addedForComparison);
-  const time = useMemo(() => `${new Date(dateTime).getHours()}:${new Date(dateTime).getMinutes()}`, [dateTime]);
-  const date = useMemo(() => new Date(dateTime).toISOString().split('T')[0], [dateTime]);
+  const time = useMemo(() => formatHour(dateTime as TDateTimeString), [dateTime]);
```
mlu-ssoft, line 84: do not use as, you can create date in `TDateTimeString` format using functions from utils

libs/shared/utils/authorization/src/lib/use-auth-client.hook.tsx
```diff
@@ -20,11 +26,26 @@ type TUseAuthProps = {
 const isUserAuthenticated = (authClient: IAuthAdapter) =>
   !!authClient.getToken().idToken && !!authClient.getToken().token;
 
+const userWorkspace = (authClient: IAuthAdapter) => {
```
mlu-ssoft, line 29: you shouldn't return variable like userWorkspace from auth library:
- this is generic library, when you will move it to another project there maybe no `userWorkspace`
- instead of returning `userWorkspace` (which is project specific) you could return `claims` (which is generic, most of auth libraries has something like claims). Claims are user specific data. Claims should be `generic`/`unknown` type

package.json
```diff
@@ -38,6 +38,7 @@
     "date-fns": "3.6.0",
     "i18next": "23.12.2",
     "i18next-http-backend": "2.5.2",
+    "jwt-decode": "^4.0.0",
```
mlu-ssoft, line 41: remove `^`

mlu-ssoft (changes requested):

PR #97, opened 2024-11-08
UKRIW-387 - add support for running multiple functions

mlu-ssoft:
Add support for running multiple functions in a Workflow.

tasks:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-387


KatarzynaPaszko (commented):
Screenshot 2024-11-08 at 09 22 39

Please look at styling of dropdown. Checkmark of selected function is on text of function name.

KatarzynaPaszko (commented):
https://github.com/user-attachments/assets/960ad6d7-c786-46c4-ad59-9b543028f1b6

First Node with functions should be disabled once there is a second Node added. It should be able for edition just after we delete Clipping function

KatarzynaPaszko (commented):
Once user open Node Function dropdown, the whole workflow should be scrolled down, so user will be able to see names in dropdown

KatarzynaPaszko (commented):
Something with Errors broke down, and they are not displayed correctly in every scenario, eg below I got 500 but I didnt see any Error Notification:

Screenshot 2024-11-08 at 09 30 33

KatarzynaPaszko (commented):
Please update storybook, so components still will be testable. Especially case with Function Nodes, where there are more than 2 Function Nodes

KatarzynaPaszko (commented):
There is a validation error on Serach Mode on Data Sets, once you select Data Sets in AC Mode

KatarzynaPaszko (approved):

PR #99, opened 2024-11-12
feat: add color map for footprint type color map

mlu-ssoft:
- add color map for footprint type coloMap
- fix error for footprint type classification with only one color


libs/map/ui-map/src/lib/true-color-image-layer/stac/classification.color-map.ts
```diff
@@ -0,0 +1,95 @@
...
+  for (let i = 0; i <= options.length; ++i) {
+    const item = options[i];
+    if (item) {
+      const color = hexToRgb(`#${item['color-hint']}`);
+
+      stops.push([['==', ['band', 1], item.value], color ? color : ['color', 0, 0, 0]]);
+    }
+  }
+
+  return stops;
```
KatarzynaPaszko, line 50:
```suggestion
  const steps: [['==', ['band', number], number], TColorMapPair][] = [];

  for (let i = 0; i <= options.length; ++i) {
    const item = options[i];
    if (item) {
      const color = hexToRgb(`#${item['color-hint']}`);

      steps.push([['==', ['band', 1], item.value], color ? color : ['color', 0, 0, 0]]);
    }
  }

  return steps;
```

libs/map/ui-map/src/lib/true-color-image-layer/stac/color-map.color-map.ts
```diff
@@ -0,0 +1,116 @@
...
+    z.literal('oxygen'),
+    z.literal('par'),
+    z.literal('phase'),
+    z.literal('salinity'),
+    z.literal('temperature'),
+    z.literal('turbidity'),
+    z.literal('velocity-blue'),
+    z.literal('velocity-green'),
+    z.literal('cubehelix'),
+  ]),
```
KatarzynaPaszko, line 50: should we add just string as well, to not get error once there will be some new color added that is not mapped here?

libs/map/ui-map/src/lib/true-color-image-layer/stac/color-map.color-map.ts
```diff
@@ -0,0 +1,116 @@
...
+    colors.reverse();
+  }
+
+  for (let i = 0; i < steps; i++) {
+    stops[i * 2] = min + i * delta;
+    stops[i * 2 + 1] = colors[i];
+  }
+
+  return stops;
+};
```
KatarzynaPaszko, line 74: should it be `stops` or `steps`  ?

libs/map/ui-map/src/lib/true-color-image-layer/use-stac-layer.hook.tsx
```diff
@@ -75,8 +75,11 @@ export const useStacLayer = () => {
           return options;
         },
       });
+      // todo remove after rewriting ol-stac library
+      setTimeout(() => {
+        handleSourceReady();
+      }, 1000);
```
KatarzynaPaszko, line 81: Why it is added ?

libs/map/ui-map/src/lib/true-color-image-layer/stac/color-map.color-map.ts
mlu-ssoft: no, library which we are using supports only those algotithms

libs/map/ui-map/src/lib/true-color-image-layer/stac/color-map.color-map.ts
mlu-ssoft: it was stops in the examples

libs/map/ui-map/src/lib/true-color-image-layer/use-stac-layer.hook.tsx
mlu-ssoft: it is hack for zooming to Footprint. when data is passwd library is firing event surceready. it is during initialization so when we listening on this event it is already fired and `handleSourceReady` won't be executed. If we add setTimeout then it works properly

libs/map/ui-map/src/lib/true-color-image-layer/stac/classification.color-map.ts
mlu-ssoft: it is `stops` in the examples

KatarzynaPaszko (approved):

PR #101, opened 2024-11-14
UKRIW-367 - added support for disabled Preset

KatarzynaPaszko:
- Added support for disabled preset
- added info about Max size of AOI in Area Node
- removed Import/Export from Workflow tab footer


mlu-ssoft (changes requested):

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/value-node.component.tsx
```diff
@@ -52,26 +46,40 @@ const formatArea = function (text: string, value: TAreaNode['value'], unit: 'km'
...
+export const ValueNode = ({ node, onClearButtonClick, error }: TValueNodeProps) => {
   const { t } = useTranslation();
   const { canActivateNode, isLast } = useActionCreator();
 
   return (
     <Node
       type={node.type}
       active={true}
-      text={formatArea(t('MAP.ACTION_CREATOR_PANEL.WORKFLOW.NODE.AREA.DESCRIPTION'), node.value, 'miles')}
+      text={formatArea(t('MAP.ACTION_CREATOR_PANEL.WORKFLOW.NODE.AREA.DESCRIPTION'), node.value, 'km')}
```
mlu-ssoft, line 78: use unit from settings

mlu-ssoft (approved):

@types/i18next/resources.d.ts
```diff
@@ -149,7 +149,7 @@ interface Resources {
             AREA: {
               TITLE: 'Area';
               DEFAULT_VALUE: 'Drawing 1';
-              INSTRUCTIONS: 'Use the drawing tools to define an area of interest';
+              INSTRUCTIONS: 'Use the drawing tools to define an area of interest. Please note we cannot return results for areas greater than 1,000m2';
```
mlu-ssoft, line 152: generate this file once again, I think this string is wrong

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/value-node.component.tsx
```diff
@@ -63,15 +75,34 @@ type TValueNodeProps = {
...
+  const { measurmentUnit, aoiLimit } = useSettings();
+  const aoiLimitInfo = useMemo(
+    () =>
+      t('MAP.ACTION_CREATOR_PANEL.WORKFLOW.NODE.AREA.ALLOWED_SIZE', {
+        maxSize: convertUnits(aoiLimit, measurmentUnit),
+      }),
+    [t, aoiLimit, measurmentUnit]
+  );
+
+  const getErrorMessage = useCallback(() => {
```
mlu-ssoft, line 87: why not `useMemo`?

mlu-ssoft (changes requested):

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/value-node.component.tsx
KatarzynaPaszko: updated

@types/i18next/resources.d.ts
KatarzynaPaszko: updated

mlu-ssoft (approved):

PR #110, opened 2024-11-26
UKRIW-427 - onboarding tooltips

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-427


libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/area-node.component.tsx
```diff
@@ -46,13 +46,23 @@ type TAreaNodeNodeProps = { node: TAreaNode };
...
-    context: { goToNextOnboardingStep, onboardingSteps },
+    context: { goToNextOnboardingStep, onboardingSteps, currentStep },
   } = useOnboarding();
   const { setActiveNode, setValue, canActivateNode } = useActionCreator();
   const { shape, setShape } = useAoi();
   const nodeRef = useRef<HTMLDivElement>(null);
   const canBeActivated = useMemo(() => canActivateNode(node), [node, canActivateNode]);
 
+  useEffect(() => {
+    if (currentStep === onboardingSteps.DRAWING_TOOLS.step_name && node.value) {
```
mlu-ssoft, line 57: I would move checking if it is current step into `goToNextOnboardingStep` function

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/area-node.component.tsx
```diff
@@ -46,13 +46,23 @@ type TAreaNodeNodeProps = { node: TAreaNode };
...
   const {
-    context: { goToNextOnboardingStep, onboardingSteps },
+    context: { goToNextOnboardingStep, onboardingSteps, currentStep },
   } = useOnboarding();
   const { setActiveNode, setValue, canActivateNode } = useActionCreator();
   const { shape, setShape } = useAoi();
   const nodeRef = useRef<HTMLDivElement>(null);
   const canBeActivated = useMemo(() => canActivateNode(node), [node, canActivateNode]);
 
+  useEffect(() => {
```
mlu-ssoft, line 56: why useEffect? shouldn't it by fired by events (eg user click)

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/area-node.component.tsx
```diff
@@ -82,9 +92,9 @@ export const AreaNode = ({ node }: TAreaNodeNodeProps) => {
     <OnboardingTooltip
       tipLocation='right'
       stepName={onboardingSteps.AREA_NODE.step_name}
-      content={onboardingSteps.AREA_NODE.tooltip_text}
-      onClick={goToNextOnboardingStep}
-      className='top-0 left-[-110px]'
+      content={onboardingSteps.AREA_NODE.tooltip_content}
+      onClick={handleGoToNextOnboardingStep}
+      reference={nodeRef}
```
mlu-ssoft, line 97: maybe `elementRef` instead of `reference`?

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/data-set/data-set-node.component.tsx
```diff
@@ -39,19 +39,26 @@ type TDataSetNodeProps = { node: TDataSetsNode };
...
-    context: { goToNextOnboardingStep, onboardingSteps },
+    context: { goToNextOnboardingStep, onboardingSteps, currentStep },
   } = useOnboarding();
   const { setActiveNode, setValue, canActivateNode } = useActionCreator();
   const { dataSet, error, updateDataSets } = useActiveDataSet();
   const nodeRef = useRef<HTMLDivElement>(null);
   const canBeActivated = useMemo(() => canActivateNode(node), [node, canActivateNode]);
 
+  const handleGoToNextOnboardingStep = useCallback(() => {
+    if (currentStep === onboardingSteps.DATA_SET_NODE.step_name) {
```
mlu-ssoft, line 50: this condition could be moved to `goToNextOnboardingStep`, I expect that it will always be checked

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/function/function-node.component.tsx
```diff
@@ -109,6 +109,19 @@ export const NodeFunction = ({ node }: IFunctionNodeProps) => {
     }
   }, [canBeActivated, node, setActiveNode]);
 
+  useEffect(() => {
+    const scrollWithTimeout = () => {
+      nodeRef.current?.scrollIntoView({ behavior: 'smooth', block: 'center' });
```
mlu-ssoft, line 114: how this `scrollIntoView` will work with the one from line 107?

libs/shared/design-system/src/lib/text/text.tsx
```diff
@@ -11,7 +11,7 @@ type TReactNode =
 interface ITextProps {
   content: ParseKeys | TReactNode;
   translate?: boolean;
-  type?: 'p' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'span';
```
mlu-ssoft, line 14: I don't think that Text element should be `div`. Those elements are used inside other elements, eg buttons, spans, it may generate wrong html semantics. if you want to element displayed as block I think better option is add styles

libs/shared/ui/ac-workflow-onboarding/src/lib/ac-workflow-onboarding.context.tsx
```diff
@@ -38,65 +40,100 @@ const translationsPath = 'MAP.ACTION_CREATOR_PANEL.ONBOARDING.STEPS';
...
       AREA_NODE: {
         step_name: 'AREA_NODE',
         next_step: 'DRAWING_TOOLS',
-        tooltip_text: t(`${translationsPath}.AREA_NODE`),
+        tooltip_content: (
+          <div>
+            <p>{t(`${translationsPath}.AREA_NODE`)}</p>
+            <Checkbox
+              label={t(`${translationsPath}.DONT_SHOW_IT_AGAIN`)}
+              // {...register('permanentHidden')}
```
mlu-ssoft, line 87: remove

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
```diff
@@ -20,34 +35,96 @@ export const OnboardingTooltip = ({
   children,
   onClick,
   className,
+  reference,
 }: PropsWithChildren<IOnboardingTooltipProps>) => {
   const [isOpen, setIsOpen] = useState(true);
+  const [positionOfHookElememnt, setPositionOfHookElememnt] = useState<TPosition>();
```
mlu-ssoft, line 41: maybe just `elementPosition?`? Hook for me is eg useEffect or useMemo, so React hooks

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
```diff
@@ -20,34 +35,96 @@ export const OnboardingTooltip = ({
...
+  const [positionOfTheTooltip, setPositionOfTheTooltip] = useState<{
```
mlu-ssoft, line 42: maybe `tooltipPosition`?

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
```diff
@@ -20,34 +35,96 @@ export const OnboardingTooltip = ({
...
+  }>();
+  const [tooltipVisible, setTooltipVisible] = useState(false);
   const {
-    context: { isOnboardingComplete, currentStep },
+    context: { isOnboardingComplete, currentStep, onboardingVisible },
   } = useOnboarding();
 
+  useEffect(() => {
+    const visible = currentStep === stepName && !!positionOfTheTooltip && onboardingVisible;
+    setTooltipVisible(!!visible);
```
mlu-ssoft, line 55: shouldn't visible be boolean type? you are using `!!positionOfTheTooltip` in conditions, so here it shouldn't be required

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
```diff
@@ -20,34 +35,96 @@ export const OnboardingTooltip = ({
...
+  }, [onClick, isOpen, tooltipVisible]);
+
+  useEffect(() => {
+    setTimeout(() => {
+      if (!reference?.current) {
+        return;
+      }
+      const nodePosition = reference.current.getBoundingClientRect();
+      setPositionOfHookElememnt(nodePosition);
+    }, 200);
```
mlu-ssoft, line 74: why 200? why not 0 or 300? what if there will be delay on old computer and rendering will take more time than 200ms?

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
```diff
@@ -20,34 +35,96 @@ export const OnboardingTooltip = ({
...
+  }, [reference, currentStep]);
 
-  return children;
+  useEffect(() => {
+    if (currentStep !== stepName || !reference) {
+      return;
+    }
+    const tooltipHook = positionOfHookElememnt;
```
mlu-ssoft, line 81: add one line before for readibility reasons

mlu-ssoft: if you will define variables on the top it will be easier to read

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
```diff
@@ -20,34 +35,96 @@ export const OnboardingTooltip = ({
...
+    } else if (currentStep === 'DATA_SET_PANEL') {
+      additionalOffset = -60;
+    }
+    switch (tipLocation) {
+      case 'top':
+        setPositionOfTheTooltip({
+          top: `${tooltipHook.y + tooltipHook.height + tooltipTipSize}px`,
+          left: `${tooltipHook.x}px`,
+        });
+        return;
```
mlu-ssoft, line 97: you aren't returning anything, I think it should be break in that case. the same in other `switch-case` statements

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
```diff
@@ -20,34 +35,96 @@ export const OnboardingTooltip = ({
...
+    if (!tooltipHook) {
+      return;
+    }
+    let additionalOffset = 0;
+    if (currentStep === 'FUNCTION_DROPDOWN') {
+      additionalOffset = 13;
+    } else if (currentStep === 'DATA_SET_PANEL') {
+      additionalOffset = -60;
+    }
+    switch (tipLocation) {
```
mlu-ssoft, line 91: add one line before for readibility reasons

mlu-ssoft (changes requested):

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/area-node.component.tsx
KatarzynaPaszko: It should appear once this input will have value, eg. once user will finish drawing a shape on map

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/data-set/data-set-node.component.tsx
KatarzynaPaszko: It is there already. I just didnt remove it from here yet.

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
KatarzynaPaszko: hmm, I think you are right. I will increase it a bit to be on the safe side. Do you think 300 or 400 is enough?

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/function/function-node.component.tsx
KatarzynaPaszko: I have cleaned it up

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
KatarzynaPaszko: I wanted to be very distinctive that it's not tooltip position, but position of an element that I hook this tooltip into, so I think its correct name

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
KatarzynaPaszko: You are right, I have updated

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/area-node.component.tsx
```diff
@@ -46,13 +46,34 @@ type TAreaNodeNodeProps = { node: TAreaNode };
...
   const { shape, setShape } = useAoi();
   const nodeRef = useRef<HTMLDivElement>(null);
   const canBeActivated = useMemo(() => canActivateNode(node), [node, canActivateNode]);
 
+  useEffect(() => {
+    if (currentStep === onboardingSteps.AREA_NODE.step_name) {
+      if (isWorkflowStarted) {
+        hideOnboardingTooltip();
+      } else {
+        showOnboardingTooltip();
```
mlu-ssoft, line 61: starting/hidding onboarding shouldn't be done in Area node. Right now logic is splitted: onboarding is started in ActionCreatorPanel and also here. it should be one place

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/function/active-node.component.tsx
```diff
@@ -22,7 +44,16 @@ export const ActiveNode = ({ node, options, onChange }: TValueNodeProps) => {
...
-      <NodeSelect value={undefined} options={options} onChange={onChange} />
+      <OnboardingTooltip
+        tipLocation='right'
+        stepName={onboardingSteps.FUNCTION_DROPDOWN.step_name}
+        content={onboardingSteps.FUNCTION_DROPDOWN.tooltip_content}
+        elementRef={nodeRef}
+      >
+        <div id={node.id} ref={nodeRef}>
+          <NodeSelect value={undefined} options={options} onChange={onChange} />
+        </div>{' '}
```
mlu-ssoft, line 55: remove `{' '}`

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/node.component.tsx
```diff
@@ -142,9 +148,16 @@ export const Node = ({
       )}
       {!hasNextNode && (
         <div className={styles.add.container}>
-          <button className={styles.add.button} disabled={!canAddNode} onClick={handleAddNode}>
-            <Icon name='Add' />
-          </button>
+          <OnboardingTooltip
```
mlu-ssoft, line 151: this component is used by other components, eg: ValueNode, AreaNode. I think onboarding will be diplicated then

libs/map/ui-search-view/src/lib/date-range-picker/date-range-picker.component.tsx
```diff
@@ -55,25 +56,24 @@ export const DateRangePicker = ({ dateMin, dateMax }: IDateRangePickerProps) =>
...
+      <div className={styles.container} ref={datePickerRef}>
+        <div className={styles.header} onClick={toggleOpen}>
+          <Text
+            content='MAP.SEARCH_VIEW.DATE_RANGE_PICKER.TITLE'
+            type='h2'
+            fontSize='large'
+            fontWeight='bold'
+            className={styles.textTitle}
+          />
+          <Icon name='ArrowDown' width={24} height={24} className={`${styles.icon} ${isOpen ? '' : 'rotate-180'}`} />
```
mlu-ssoft, line 74: maybe move this into styles property?: `styles.icon(isOpen)`

libs/shared/design-system/src/lib/input/checkbox/checkbox.tsx
```diff
@@ -30,12 +30,24 @@ export interface ICheckboxProps {
...
+    {
+      id,
+      onChange,
+      onBlur,
+      label,
+      name,
+      state,
+      disabled,
+      className = '',
+      labelClassName,
```
mlu-ssoft, line 48: set default values otherwise it will be parsed to undefined

libs/shared/design-system/src/lib/tooltip/tooltip.tsx
```diff
@@ -9,12 +9,20 @@ type TTipLocation = 'top' | 'bottom' | 'left' | 'right';
 
 interface ITooltipProps {
   tipLocation: TTipLocation;
-  content: string;
+  content: string | JSX.Element;
   className?: string;
   isOpen?: boolean;
+  tooltipPosition?: React.CSSProperties;
```
mlu-ssoft, line 15: whats the difference between position and location?

libs/shared/ui/ac-workflow-onboarding/src/lib/ac-workflow-onboarding.context.tsx
```diff
@@ -38,72 +43,116 @@ const translationsPath = 'MAP.ACTION_CREATOR_PANEL.ONBOARDING.STEPS';
 
 const OnboardingContext = createContext<IOnboardingContextType | undefined>(undefined);
 
-export const OnboardingProvider: FC<{ children: ReactNode }> = ({ children }) => {
-  const [currentStep, setCurrentStep] = useState<TStepName>('NOT_STARTED');
-  const [displayOnboardingModal, setDisplayOnboardingModal] = useState(false);
+export const OnboardingProvider = ({ children }: PropsWithChildren) => {
+  const [currentStep, setCurrentStep] = useState<TStepName>('AREA_NODE');
```
mlu-ssoft, line 47: it is better to pass this as a `props` in `OnboardingProvider`, eg `defaultStep`/`firstStep`. Then it won't be hardcoded in the library

libs/shared/ui/ac-workflow-onboarding/src/lib/ac-workflow-onboarding.context.tsx
```diff
@@ -38,72 +43,116 @@ const translationsPath = 'MAP.ACTION_CREATOR_PANEL.ONBOARDING.STEPS';
...
       FUNCTION_DROPDOWN: {
         step_name: 'FUNCTION_DROPDOWN',
+        next_step: 'ADD_FUNCTION_NODE',
+        tooltip_content: t(`${translationsPath}.FUNCTION_DROPDOWN`),
+      },
+      ADD_FUNCTION_NODE: {
+        step_name: 'ADD_FUNCTION_NODE',
         next_step: 'FINISH',
-        tooltip_text: t(`${translationsPath}.FUNCTION_DROPDOWN`),
+        tooltip_content: t(`${translationsPath}.ADD_FUNCTION_NODE`),
```
mlu-ssoft, line 123: why snake case? we are using camel case everywhere. Snake case is used for capitalize letters only

libs/shared/ui/ac-workflow-onboarding/src/lib/ac-workflow-onboarding.context.tsx
```diff
@@ -38,72 +43,116 @@ const translationsPath = 'MAP.ACTION_CREATOR_PANEL.ONBOARDING.STEPS';
...
 
-  const updateShouldDisplayOnboardingModal = useCallback(
-    (collapsed?: boolean, enabled?: boolean) => {
-      return !isOnboardingComplete && currentStep === 'NOT_STARTED' && !collapsed && !enabled;
-    },
-    [currentStep, isOnboardingComplete]
-  );
-
-  const goToNextOnboardingStep = () => {
+  const goToNextOnboardingStep = (currentTooltipStep: TStepName) => {
```
mlu-ssoft, line 129: use `useCallback`. this is provider so any change in here will affect every child, whole app in this case

libs/shared/ui/ac-workflow-onboarding/src/lib/ac-workflow-onboarding.context.tsx
```diff
@@ -38,72 +43,116 @@ const translationsPath = 'MAP.ACTION_CREATOR_PANEL.ONBOARDING.STEPS';
...
       completeOnboarding();
     } else {
-      setCurrentStep(onboardingSteps[currentStep].next_step as TStepName);
+      if (currentTooltipStep === currentStep) {
+        setCurrentStep(onboardingSteps[currentStep].next_step as TStepName);
+      }
     }
   };
 
   const completeOnboarding = () => {
```
mlu-ssoft, line 146: use `useCallback`. this is provider so any change in here will affect every child, whole app in this case

libs/shared/ui/ac-workflow-onboarding/src/lib/ac-workflow-onboarding.context.tsx
```diff
@@ -38,72 +43,116 @@ const translationsPath = 'MAP.ACTION_CREATOR_PANEL.ONBOARDING.STEPS';
...
     setIsOnboardingComplete(true);
   };
 
+  const resetOnboarding = () => {
```
mlu-ssoft, line 150: use `useCallback`. this is provider so any change in here will affect every child, whole app in this case

libs/shared/ui/ac-workflow-onboarding/src/lib/ac-workflow-onboarding.context.tsx
```diff
@@ -38,72 +43,116 @@ const translationsPath = 'MAP.ACTION_CREATOR_PANEL.ONBOARDING.STEPS';
...
-  const goToNextOnboardingStep = () => {
+  const goToNextOnboardingStep = (currentTooltipStep: TStepName) => {
+    if (isOnboardingComplete || permanentHidden) {
+      return;
+    }
+    if (dontShowAgain) {
+      handleHidePermanently();
+      completeOnboarding();
+    }
     if (currentStep === 'FINISH') {
```
mlu-ssoft, line 137: add one line before for readibility reason

libs/shared/ui/ac-workflow-onboarding/src/lib/ac-workflow-onboarding.context.tsx
```diff
@@ -38,72 +43,116 @@ const translationsPath = 'MAP.ACTION_CREATOR_PANEL.ONBOARDING.STEPS';
...
-    },
-    [currentStep, isOnboardingComplete]
-  );
-
-  const goToNextOnboardingStep = () => {
+  const goToNextOnboardingStep = (currentTooltipStep: TStepName) => {
+    if (isOnboardingComplete || permanentHidden) {
+      return;
+    }
+    if (dontShowAgain) {
```
mlu-ssoft, line 133: add one line before for readibility reason

libs/shared/ui/ac-workflow-onboarding/src/lib/ac-workflow-onboarding.context.tsx
```diff
@@ -38,72 +43,116 @@ const translationsPath = 'MAP.ACTION_CREATOR_PANEL.ONBOARDING.STEPS';
...
+      handleHidePermanently();
+      completeOnboarding();
+    }
     if (currentStep === 'FINISH') {
       completeOnboarding();
     } else {
-      setCurrentStep(onboardingSteps[currentStep].next_step as TStepName);
+      if (currentTooltipStep === currentStep) {
```
mlu-ssoft, line 140: if you will add return in first if then you don't need to use else and you will reduce one level of nesting

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
```diff
@@ -1,54 +1,154 @@
...
+        additionalOffset = -60;
+
+        break;
+      }
+      case 'ADD_FUNCTION_NODE': {
+        additionalOffset = 25;
+
+        break;
+      }
+      // No default
```
mlu-ssoft, line 106: I see that there is no default. This comment isn't adding any additional value. comments make sense if they are explaining something that can be understand from the code. if you would like to handle only three cases (and there is more) it is better to use if-else statement. then you don't need to add comment here.
try to answer this: in a year or two will you able to explain this comment? why it was added? why are you supported only three cases? how much time you will need to figure it out? code should be self explanatory

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
mlu-ssoft: so maybe move this into `usePosition` hook? then it will be easier to read if code if it will be in isolation

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
mlu-ssoft: as I understand this is position of `elementRef`? is so then maybe `positionElementRef`? I know that this variable doesn't sound beautifull but you can guess what it is. Hook in context of react has different meaning and I think that a lot of people will associate it with that

mlu-ssoft (changes requested):

libs/shared/ui/ac-workflow-onboarding/src/lib/onboarding-tooltip/onboarding-tooltip.tsx
mlu-ssoft: I would leave it for now, I will add an comment to fix it later using Popper component from Material, it will solve problem

mlu-ssoft (approved):

PR #111, opened 2024-11-27
feat: add notifications for workflows

mlu-ssoft:
- add green dot when notification is finished (ready or error):
* use Long Pull technique (there is request made in background on every 10 seconds if there are workflows to process)
* add optimistic update for workflows. Cache is invalidated in cases
> user openned workflow history tab
> there was proccessed workflow (ready or failed status)
> after creating new workflow (if it wasn't then the user after going to History tab will see old results for few seconds until new results won't be donloaded). I didn't implemented optimistic update as it is hard to do this in current implementation - there is pagination based on params
- fix bug when switching to another tab while workflow was in progress (InProgressWorkflow/FinishedWorkflow modal dissapear while should be visible)

Tickets:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-117


KatarzynaPaszko (commented):
It does not work as expected:
- Once Im on History tab, green dot does not appear once workflow is ready
Screenshot 2024-11-27 at 15 03 38

- Once on different tab, green dot appear with delay 5-8 minutes after workflow is ready
image

PR #112, opened 2024-12-04
UKRIW-118 - import and export workflow configuration

mlu-ssoft:
Add import/export workflow configuration. Support only version 1.0 for now.

Tickets:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-118


KatarzynaPaszko (commented):
https://github.com/user-attachments/assets/bd2ac2ab-2c45-438f-a20a-c6df2e9ff4c4

I think import/export button should be disabled once there is any modal displayed in Workflow tab

KatarzynaPaszko (commented):
According to AC5, once there is an error with loading JSON file, and error message should appear on the bottom of the page. I have modified JSON file to be broken, and tried to upload it, but error message didnt appear. I just see error in console:
Screenshot 2024-12-04 at 14 03 03

libs/map/data-access-map/src/lib/store/action-creator/workflow-import/import-workflow.schema.ts
```diff
@@ -0,0 +1,112 @@
...
+  value: z
+    .union([
+      z.literal('sentinel-1'),
+      z.literal('sentinel-2-l1c'),
+      z.literal('sentinel-2-l2a'),
+      z.literal('sentinel-3'),
+      z.literal('sentinel-5p'),
+      z.literal('esacci-globallc'),
+      z.literal('clms-corinelc'),
+      z.literal('clms-water-bodies'),
```
KatarzynaPaszko, line 47: should we make it more generic, so it will work with new data sets that will be added in the future ?
Or maye be can extract it to some separate place, and then import everywhere it is needed

libs/map/data-access-map/src/lib/store/action-creator/workflow-import/import.workflow.ts
```diff
@@ -0,0 +1,132 @@
...
+          const isActive = dateRangeImportedNodeHasValues || functionImportedNodesHasValues;
+          return { ...node, state: isActive ? 'not-active' : 'initial' };
+        }
+
+        default: {
+          return undefined;
+        }
+      }
+    })
+    .filter((item) => !!item);
```
KatarzynaPaszko, line 96: I would extract this logic to separate function
```
const getNodeState = (
  nodeType: string,
  aoiImportedNodeHasValues: boolean,
  dataSetImportedNodeHasValues: boolean,
  dateRangeImportedNodeHasValues: boolean,
  functionImportedNodesHasValues: boolean
) => {
  switch (nodeType) {
    case 'area':
      return aoiImportedNodeHasValues || dataSetImportedNodeHasValues || dateRangeImportedNodeHasValues || functionImportedNodesHasValues
        ? 'not-active'
        : 'initial';
    case 'dataSet':
      return dataSetImportedNodeHasValues || dateRangeImportedNodeHasValues || functionImportedNodesHasValues
        ? 'not-active'
        : 'initial';
    case 'dateRange':
      return dateRangeImportedNodeHasValues || functionImportedNodesHasValues
        ? 'not-active'
        : 'initial';
    default:
      return undefined;
  }
};
```
and then just reuse it here
```
 .map((node) => ({
      ...node,
      state: getNodeState(
        node.type,
        aoiImportedNodeHasValues,
        dataSetImportedNodeHasValues,
        dateRangeImportedNodeHasValues,
        functionImportedNodesHasValues
      ),
    }))
```
then it would be easier to digest what's going on here

libs/map/data-access-map/src/lib/store/action-creator/workflow-import/import-workflow.schema.ts
mlu-ssoft: right now its not fully dynamic - there are hardcoded values in few places, I wasn't able to do whole mechanism dynamic, the bigger problem is how to map dynamic values into query for AC and SEARCH. I don't know how to do this ATM and until it will be done I think it is better to hardcode it. I would like to have fully dynamic mechanism, but it is very hard to create it (and not overconfigure it). I will think how to do one type (do not duplicate it)

libs/map/data-access-map/src/lib/store/action-creator/workflow-import/import.workflow.ts
```diff
@@ -0,0 +1,153 @@
...
+        case 'area':
+        case 'dataSet':
+        case 'dateRange': {
+          return { ...node, state: getNodeState(node, importedNodes) };
+        }
+
+        default: {
+          return undefined;
+        }
+      }
```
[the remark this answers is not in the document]
mlu-ssoft: its needed, I'm handling only 3 cases (`area`, `dataSet`, `dateRange`), but not `function`. This is done on purpose, function is handled separatelly. If I don't handle this case then eslint is throwind a warning

libs/map/data-access-map/src/lib/store/action-creator/workflow-import/export.workflow.ts
```diff
@@ -0,0 +1,55 @@
...
+            order: node.order,
+            type: node.type,
+            value: node.value,
+          };
+        }
+
+        return undefined;
+      })
+      .filter((node): node is TNodeImport => !!node),
+  };
```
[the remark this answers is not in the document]
mlu-ssoft: this won't work because of typescript, I tried to simplify this but TS is throwing an error that I'm trying to assign different types (eg `area` and `dataSet`)

KatarzynaPaszko (approved):

PR #113, opened 2024-12-04
UKRIW-125 - comparison tool component

KatarzynaPaszko:
Ticket: https://ukri-spyrosoft.atlassian.net/browse/UKRIW-125

- Created Comparison Tool component and added it in Top Bar
- Created Comparison Mode store
- Disabling many features scroll whole app, once app is n ComparisonMode
- Moved all modals to one place to unify styling


libs/map/ui-search-view/src/lib/dynamic-tree-form/tree.component.tsx
```diff
@@ -30,7 +32,7 @@ export const DynamicTreeForm = ({ tree }: TTreeProps) => {
       >
         <div ref={treeRef}>
           {treeBuilder.items.map((item) => (
-            <TreeElement key={item.id} item={item} disabled={disabled} />
+            <TreeElement key={item.id} item={item} disabled={comparisonMode || disabled} />
```
mlu-ssoft, line 35: the same situation as in DateRange component, I would merge `comparisonMode || disabled` into one `disabled` variable

mlu-ssoft (changes requested):

libs/map/data-access-map/src/lib/store/comparison-mode/comparison-tool.store.tsx
```diff
@@ -0,0 +1,95 @@
...
+  id: string;
+};
+
+interface IComparisonToolStore {
+  comparisonItems: {
+    firsItemId?: string;
+    secondItemId?: string;
+    items: TComparisonItem[];
+  };
+  comparisonMode: boolean;
```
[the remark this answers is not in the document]
KatarzynaPaszko: Im using it in multiple places across app. I want it to be self explanatory on first glance what it means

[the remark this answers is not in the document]
mlu-ssoft: but it isn't, at least for me. what `comparisonMode` means? is it state or status? for me it looks like a state/enum. `comparisonModeEnabled` will be better name

also you can't read name without context. context should be treated as a prefix, the whole. take a look on libraries how they are doing it:
- `mui` modal, there is variable `isOpen`, not `isModalOpen` or `isMuiModalOpen`
- react hook forms: `handleSubmit`, not `handleSubmitForm`. Control, not FormControl. Only exception here is that there are two state values (formState and getFieldState), but because there are two different values there in the library
- angular material: `<mat-tab label="First">`, `label` not `matTabLabel`
- angular is wrapping native inputs with directives so you can use html, eg add `disabled` attribute to it, not `angularDisabled`/`inputDisabled`
we can have more specific names if you want, but I'm in a state that code should be readable as a story, like human language. and in human language you have context, you aren't saying `yesterday I was in a camp, in that camp there was camp shower and a lot of camp tents. there were a lot of camp people`. You would just say: `I was in a camp, there was shower and lot of tents. There were a lot of people`. Code should be from one point readable and usable for developers, but from another point should tell a story, eg should explain business purpose, language etc. It should be possible to explain what it does from business perspective by just reading the code as a sentece. I undestand that namings should be self explanatory, but I'm treating context a part of the naming. I don't have problem with that, we can add prefixes, I just want to point out my thoughs on that topic

libs/map/feature-timeline-analytics-dashboard/src/lib/timeline-analytics-dashboard.component.tsx
```diff
@@ -8,6 +10,9 @@ type TActionCreatorPanelProps = {
 
 export const TimelineAnalyticsDashboard = ({ className }: TActionCreatorPanelProps) => {
   const { status, minDate, maxDate, selectedMinDate, selectedMaxDate, updateSelectedDate } = useTimelineAnalytics();
+  const { comparisonMode } = useComparisonMode();
+
+  const isDisabled = useMemo(() => comparisonMode || status === 'pending', [comparisonMode, status]);
```
mlu-ssoft, line 15: maybe just `disabled`? this is the name used in other places

libs/map/feature-comparison-tool/src/lib/comparison-tool.component.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+interface IComparisonToolProps {
+  className?: string;
+}
+
+export const ComparisonTool = ({ className }: IComparisonToolProps) => {
+  const { t } = useTranslation();
+  const { comparisonMode, comparisonItems, toggleComparisonMode, removeComparisonItem } = useComparisonMode();
+
+  const isDisabled = useMemo(() => !(comparisonItems.firsItemId && comparisonItems.secondItemId), [comparisonItems]);
```
mlu-ssoft, line 17: maybe just `disabled`? this is the name used in other places

libs/map/ui-search-view/src/lib/search-view.component.tsx
```diff
@@ -73,6 +75,8 @@ export const SearchView = ({
     setMode(schema);
   }, [schema, setMode]);
 
+  const isDisabled = useMemo(() => comparisonMode || !form.formState.isValid, [comparisonMode, form.formState.isValid]);
```
mlu-ssoft, line 78: maybe just `disabled`? this is the name used in other places

libs/map/data-access-map/src/lib/store/comparison-mode/comparison-tool.store.tsx
```diff
@@ -0,0 +1,103 @@
...
+  }))
+);
+
+export const useComparisonMode = () => {
+  const comparisonState = useComparisonToolStore((state) => ({
+    comparisonItems: state.comparisonItems,
+    comparisonMode: state.comparisonMode,
+  }));
+
+  const toggleComparisonMode = useComparisonToolStore((state) => state.toggleComparisonMode);
```
mlu-ssoft, line 84: you will have a lot of actions fired, add console log and look on the devtools. I would return one hook `useComparisonToolStore(state => mapValuesHere)`. otherwise use `useMemo` (but check if wont cause rerenders anyway)

mlu-ssoft (changes requested):

mlu-ssoft (approved):

PR #143, opened 2025-01-13
[UKRIW-668] disable buttons in AC when AC is not enabled

mlu-ssoft:
- disable buttons in AC when AC is not enabled.
- not load supported data sets when AC is not enabled

Tickets:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-668


libs/map/feature-action-creator-panel/src/lib/content/workflow/workflow.component.tsx
KatarzynaPaszko, line 1: Should we also not make a call for functions if AC is not enabled?

mlu-ssoft: we can call it - sooner the better, if user will activate AC then functions will be already downloaded

KatarzynaPaszko (approved):

PR #145, opened 2025-01-16
UKRIW-564 - bug fix - Can run the workflow for an area larger than 1000 km2

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-564

Bug fix

### DIFF
--- a/libs/map/feature-action-creator-panel/src/lib/content/workflow/workflow.component.tsx
+++ b/libs/map/feature-action-creator-panel/src/lib/content/workflow/workflow.component.tsx
@@ -11,7 +11,8 @@ import {
   useFunctions,
 } from '@ukri/map/data-access-map';
 import { Button } from '@ukri/shared/design-system';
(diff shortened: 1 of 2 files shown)


libs/map/feature-action-creator-panel/src/lib/content/workflow/workflow.component.tsx
```diff
@@ -93,6 +96,14 @@ export const Workflow = () => {
     setSupportedDataSets(options);
   }, [data, isFunctionsLoaded, enabled, setSupportedDataSets]);
 
+  const isAreaIncorrect = useMemo<boolean>(() => {
```
mlu-ssoft, line 99: can you reverse condition? usually it is checked if something is true, not false, so `isCorrectArea` or `checkAreaSize`

mlu-ssoft (changes requested):

mlu-ssoft (approved):

PR #148, opened 2025-01-22
[UKRIW-56] [UKRIW-59] add measure distance

mlu-ssoft:
Add support for measure distance:
- switching between drawing polygon and line (in measure distance)
- unify utils related to geometry in `data-access-map`
- move area and distance calculations into `data-access-map`
- calculate area and distance basing on units (km/km2/miles/miles2)
- change Toogle component to support two types: toggle and switch (toggle is on/off, switch can switch between values)
- add measure distance panel and layer
- group code related to layers in `layers` directory 

Tickets:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-56
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-59


KatarzynaPaszko (commented):
https://github.com/user-attachments/assets/3cba4fe8-2519-4ff9-9dc0-3aea449940e1

measurement tool UI is shaking/jumping once length is changing. Please fix it

KatarzynaPaszko (changes requested):
https://github.com/user-attachments/assets/d23cda26-9bdf-4580-9de9-4c24eba19cc0

Once I change measurement for miles, it initially shows miles but then changes back to kilometers

KatarzynaPaszko (commented):
https://github.com/user-attachments/assets/e0c4a62b-fa03-4d16-89fc-48939e5cfe7a

- UI is jumping a lot, once you change tool. Please discuss with Toby if he can come up with solution that is more static, so it will be more usable (especially on mobile)
- Why in order to change measurement tool user use switch, and not click on icon itself. That would be more intuitive (like in other systems) and in case there will be more tools in the future, then there wont be an issue with adding it.

KatarzynaPaszko (commented):
Screenshot 2025-01-23 at 09 20 44

On implementation you have `miles` and `miles2`, but on design it says `m` and` m2`
Screenshot 2025-01-23 at 09 20 55

libs/map/data-access-map/src/lib/geometry/area-and-distance.ts
```diff
@@ -0,0 +1,118 @@
+import { LineString } from 'ol/geom';
+import { getArea as getAreaFromGeometry } from 'ol/sphere';
+
+import { createGeometry } from './geometry';
+import { TCoordinate, TUnit, TUnitType } from './shape.model';
+
+const SQUARE_KM_TO_SQUARE_MILE = 0.386102;
+const KM_TO_MILE = 0.621371192;
```
KatarzynaPaszko, line 8: Why `SQUARE_KM_TO_SQUARE_MILE` has 6 digits after coma and `KM_TO_MILE` has 9 digits after coma?

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/value-node.component.tsx
```diff
@@ -29,46 +22,24 @@ const getIconFromShape = (value: TAreaNode['value']): 'Polygon' | 'Circle' | 'Sq
...
-      const value = Math.round((area / 1000000) * 100) / 100;
-
-      output = `${value} miles<sup>2</sup>`;
-      break;
+export const formatUnit = (area: number, unit: TBaseUnit) => {
+  const value = convertUnits(area, convertBaseUnitToAreaUnit(unit));
+  switch (value.unit.type) {
+    case 'km2':
+    case 'miles2': {
+      return `${value.value} ${value.unit.displayedValue}<sup>2</sup>`;
```
KatarzynaPaszko, line 30: Please add `displayedValue` to translations

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/value-node.component.tsx
```diff
@@ -29,46 +22,24 @@ const getIconFromShape = (value: TAreaNode['value']): 'Polygon' | 'Circle' | 'Sq
...
-        const value = Math.round((area / 1000000) * 100) / 100;
-        output = `${value} km<sup>2</sup>`;
-      } else {
-        const value = Math.round(area * 100) / 100;
-        output = `${value} m<sup>2</sup>`;
-      }
-      break;
+    case 'km':
+    case 'miles': {
+      return `${value.value} ${value.unit.displayedValue}`;
```
KatarzynaPaszko, line 35: Same here. Please add `displayedValue` to translations

libs/map/data-access-map/src/lib/geometry/area-and-distance.ts
```diff
@@ -0,0 +1,118 @@
...
+        };
+      }
+
+      return {
+        value: Math.round(value * 100) / 100,
+        unit: {
+          type: unit,
+          displayedValue: 'km',
+        },
+      };
```
KatarzynaPaszko: https://github.com/user-attachments/assets/2331e4eb-e8ae-47ea-b3fb-81ae7dcc4faa

This `if` is causing measurement issues. At very small distances it starts from 1 and gos to 100 and then changes to zero point values again as on video. 

Same issue is with` 'km2'`

libs/map/data-access-map/src/lib/geometry/area-and-distance.ts
mlu-ssoft: I don't know. Those are static values, if you search for km2 to mile2 or km to mile then you will got those values

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/value-node.component.tsx
mlu-ssoft: fixed

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/value-node.component.tsx
mlu-ssoft: fixed

libs/map/data-access-map/src/lib/geometry/area-and-distance.ts
mlu-ssoft: fixed

KatarzynaPaszko (commented):
https://github.com/user-attachments/assets/9d2b4d38-45dd-4846-bb84-a96f287e48d5

UI is still jumping

KatarzynaPaszko (commented):
Value with m2 is moving 

https://github.com/user-attachments/assets/dd0843ea-2ea4-4a32-a34a-f6b0edc2a98e

KatarzynaPaszko (commented):
https://github.com/user-attachments/assets/382be178-af5a-4886-934f-e6ad4c16f48f

UI is still shaking on miles on small values. Do not share on bigger values

KatarzynaPaszko (commented):
I can't finish drawing shape by clicking on starting point. It seems that now the only way to finish drawing shape is by clicking escape button on keyboard

https://github.com/user-attachments/assets/571b7957-2de3-496e-b4a4-70cc27d803f3

KatarzynaPaszko (commented):
Once I enter page/refresh page measurement drawing tool is enabled by default, and I am able o draw lines on map. Once I draw lines, the measurement tool tile do not show up and does not show the length of drawn shape.  
Desired behavior should be that user can start drawing lines just after clicking measurement tool button pon top menu. Otherwise it should not be possible to draw on map

KatarzynaPaszko (commented):
https://github.com/user-attachments/assets/8c35b1a1-1553-4fb5-af0b-2a2ae519ff39

Drawing tool dot is constantly on screen, even if drawing tool is disabled. As on video I have activated comparison tool, but this dot is always hanging around the view

KatarzynaPaszko (commented):
Screenshot 2025-01-27 at 10 26 43

In current state user can't see measurement tool on smaller screen. I think measurement tool component should be above AC component, because user can hide measurement tool, but not AC component.

libs/map/data-access-map/src/lib/geometry/area-and-distance.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  return getAreaFromGeometry(shape) || 0;
+};
+
+export const getLineLength = (coordinates: TCoordinate | undefined) => {
+  const shape = createGeometry(coordinates);
+
+  if (!shape || !coordinates) {
+    return 0;
+  }
```
KatarzynaPaszko, line 24: shape const is never used, just in this if statement, so no need to create it

libs/map/data-access-map/src/lib/geometry/area-and-distance.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  switch (coordinates.type) {
+    case 'circle': {
+      return 2 * Math.PI * coordinates.radius;
+    }
```
KatarzynaPaszko, line 29: we can't draw circle with measurement drawing tool, so no need to add it

mlu-ssoft: it's a generic function (I unified types for all geometry). For now we don't support measurements in circle but I would leave it to have unify interface - if we can pass circle geometry (and we can, because of unified interface) then it would be good if circle will be supported

KatarzynaPaszko (approved):

PR #165, opened 2025-01-30
UKRIW-645 - bug fix for do not show again checkbox on onboarding tooltip

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-645

### DIFF
--- a/libs/shared/design-system/src/lib/tooltip/tooltip.tsx
+++ b/libs/shared/design-system/src/lib/tooltip/tooltip.tsx
@@ -13,6 +13,7 @@ interface ITooltipProps {
   className?: string;
   isOpen?: boolean;
   tooltipPosition?: React.CSSProperties;
+  onClose?: () => void;
 }
 
 export const Tooltip = ({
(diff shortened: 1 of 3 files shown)


libs/shared/design-system/src/lib/tooltip/tooltip.tsx
```diff
@@ -32,8 +34,9 @@ export const Tooltip = ({
   const handleClose = useCallback(() => {
     if (isTooltipOpen) {
       setIsOpen(false);
+      onClose?.();
```
mlu-ssoft, line 37: how it will be compilled? is it the same as `if (onClose) { onClose() }`?

mlu-ssoft (approved):

libs/shared/design-system/src/lib/tooltip/tooltip.tsx
KatarzynaPaszko:
```
if (onClose) {
    onClose();
}
```

PR #169, opened 2025-02-07
UKRIW-636 - Selecting from assets to view or compare when one item contains several assets

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-636

- Added component for displaying multiple indices within one result item component
- Adjusted how map layer is created once we pass specific asset key 
- Adjusted comparison tool logic to support multiple indices within one item


libs/map/ui-results-view/src/lib/results-list/use-result.hook.ts
```diff
@@ -1,24 +1,36 @@
...
+  const isSelected = useCallback((id: string) => visibleFeature?.id === id, [visibleFeature]);
+  const isSelectedMultipleIndices = useCallback(
+    (id: string, key: TAssetKey) => visibleFeature?.id === id && visibleKey === key,
+    [visibleFeature, visibleKey]
+  );
 
   const isAddedToComparison = useCallback(
-    (item: TFeature) => itemAddedToComparisonMode(item),
+    (item: TFeature, key?: TAssetKey) => {
+      // console.log('isAddedToComparison itemAddedToComparisonMode(item, key)', itemAddedToComparisonMode(item, key));
```
mlu-ssoft, line 26: remove

mlu-ssoft (changes requested):

libs/map/ui-results-view/src/lib/results-list/result-item/multiple-items-action-buttons/asset-item.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+        onClick={onComparisonToggle}
+        className={addedForComparison ? '!text-error' : ''}
+        disabled={canCompare}
+      />
+      <Button
+        text={isSelected ? `${translationPath}.BUTTON_HIDE` : `${translationPath}.BUTTON_SHOW`}
+        size='small'
+        onClick={onToggleView}
+        disabled={comparisonEnabled}
+        className='ml-1 w-[41px]'
```
mlu-ssoft, line 54: why button has fixed width? shouldn't it take as much space as is available?

libs/map/ui-results-view/src/lib/results-list/result-item/multiple-items-action-buttons/multiple-items-action-buttons.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const isItemSelected = useCallback(
+    (key: TAssetKey) => {
+      return selectedIndice === key && isSelected(featureId);
+    },
+    [selectedIndice, isSelected, featureId]
+  );
+
+  useEffect(() => {
+    setItemsInComparison((prev) => prev.filter((item) => addedForComparison(item as TAssetKey)));
```
mlu-ssoft, line 69: this effect is fired on every change in comparisonItems but logic is hard to guess (why this is happening). I would add function in component store which will return true/false if item is in the stock in comparison store. If you add this function into dependency it will be more visible why it is there and effect will be the same - it will fire every time something change in component store

libs/map/ui-results-view/src/lib/results-list/result-item/multiple-items-action-buttons/multiple-items-action-buttons.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+  }, [comparisonItems, addedForComparison]);
+
+  const onComparisonToggle = useCallback(
+    (key: TAssetKey) => {
+      setItemsInComparison((prev) => (prev.includes(key) ? prev.filter((item) => item !== key) : [...prev, key]));
```
mlu-ssoft, line 74: you can use Set to make uniqueness in array: `[...new Set(myArray)]`

libs/map/ui-results-view/src/lib/results-list/results-list.component.tsx
```diff
@@ -23,30 +24,55 @@ export const ResultsList = ({ features }: IResultsListProps) => {
...
-          gridCode={feature.properties['grid:code']}
-          cloudCoverage={feature.properties['eo:cloud_cover']}
-          collectionName={feature.collection}
-          dateTime={feature.properties.datetime}
-          selected={isSelected(feature.id)}
-          onToggleSelectedItem={() => toggleItem(feature)}
-        >
-          <ActionButtons
+      {features.map((feature: TFeature) =>
+        mode === 'action-creator' && Object.keys(feature.assets).length > 1 ? (
```
mlu-ssoft, line 28: you can move this logic into `toggleItem` function. It will simplify - right now there is complex condition and only one property is different in <ResultItem. It is hard to guess why this is happening

libs/map/ui-results-view/src/lib/results-list/results-list.component.tsx
```diff
@@ -23,30 +24,55 @@ export const ResultsList = ({ features }: IResultsListProps) => {
...
+            key={feature.id}
+            className='mb-4'
+            imageUrl={feature.assets.thumbnail?.href || ''}
+            gridCode={feature.properties['grid:code']}
+            cloudCoverage={feature.properties['eo:cloud_cover']}
+            collectionName={feature.collection}
+            dateTime={feature.properties.datetime}
+            selected={isSelected(feature.id)}
+          >
+            <MultipleItemsActionButtons
```
mlu-ssoft, line 39: I would create one `ActionButtons` component which will have logic to switch between multiple and single buttons component. It will be easier to understand what it is, what it is doing and why it is here

libs/map/ui-results-view/src/lib/results-list/use-result.hook.ts
```diff
@@ -27,16 +39,21 @@ export const useResult = () => {
   }, []);
 
   const handleSelectedItemToggle = useCallback(
-    (item: TFeature) => {
-      const newFeature = selectedFeature?.id !== item.id ? item : undefined;
-      setFeature(newFeature);
+    (item: TFeature, key?: TAssetKey) => {
+      if (key) {
```
mlu-ssoft, line 43: I think it can be simplified:
const getFeature = () => {
if (key) {
return visibleFeature?.id === item.id && visibleKey === key ? undefined : item;
}

return visibleFeature?.id !== item.id ? item : undefined;
}

setFeature(getFeature(), key);

if key will be undefined then undefined will be passed - the same as in setFeature(newFeature);

mlu-ssoft (changes requested):

libs/map/data-access-map/src/lib/store/true-color-image/true-color-image.store.tsx
```diff
@@ -10,17 +10,20 @@ export const useTrueColorImageStore = create<ITrueImageStore>()(
   devtools((set) => ({
     stacUrl: undefined,
     feature: undefined,
-    setFeature: (feature: TFeature | undefined) =>
+    visibleKey: undefined,
```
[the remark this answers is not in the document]
KatarzynaPaszko: This asset have few indices (few different asset names),. Here I wanted to emphasize that we need to pass only the one we want to be visible on created layer. I wanted this name to be self descriptive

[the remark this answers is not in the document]
KatarzynaPaszko: type TAssetKey is a collection of strings "thumbnail" | "visual" | "cdom" | "cya_cells" | "doc" | "turb". I wanted to be specific what keys can be used here, I do not want it to be just any string

libs/map/ui-map/src/lib/stac/use-stac-layer-creation.tsx
```diff
@@ -96,9 +97,19 @@ export const useStacLayerCreation = () => {
...
+    async ({
+      url,
+      zIndex,
+      authorized,
+      visibleKey,
+    }: {
+      url: string;
+      zIndex: number;
+      authorized: boolean;
+      visibleKey?: string;
```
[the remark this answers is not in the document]
KatarzynaPaszko: To create layer for displaying image from water quality we do not want here to pass an array of keys. We want just one key. If we will pass multiple keys, then just one will be visible anyway, and cover others, and we do not want that.

libs/map/ui-results-view/src/lib/results-list/results-list.component.tsx
KatarzynaPaszko: I think we should keep logic here, as ResultItem also get different props depending if we display Item with one image or many indices(images)

libs/map/ui-results-view/src/lib/results-list/result-item/multiple-items-action-buttons/asset-item.component.tsx
KatarzynaPaszko: on click text change from "View" to "Hide". Those 2 words have different width, so button width change on click and therefore UI moves a little on each click. It was set to precise value in order to prevent UI jumping on click

libs/map/ui-results-view/src/lib/results-list/result-item/multiple-items-action-buttons/multiple-items-action-buttons.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+  return (
+    <div>
+      <div className='flex space-between'>
+        {canDownload && <DownloadButton onDownload={onDownload} disabled={comparisonEnabled} />}
+        <ToggleAssetsButton isOpened={isOpened} onToggle={onToggleShowAssets} itemsInComparison={itemsInComparison} />
+      </div>
+
+      {isOpened && (
+        <div className='mt-2 border-t border-t-bright-dark'>
+          {(Object.keys(indices) as TAssetKey[]).map((key) => (
```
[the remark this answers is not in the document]
KatarzynaPaszko: But I want it to be typed everywhere below as TAssetKey

libs/map/ui-results-view/src/lib/results-list/results-list.component.tsx
```diff
@@ -1,10 +1,51 @@
...
+}: IActionButtons) => {
+  if (mode === 'action-creator' && Object.keys(feature.assets).length > 1) {
+    return <MultipleItemsActionButtons feature={feature} canDownload={mode === 'action-creator'} />;
+  }
+  return (
+    <SingleItemActionButtons
+      selected={isSelected(feature.id)}
+      comparisonEnabled={comparisonEnabled}
+      addedForComparison={isItemAddedToComparisonMode(feature)}
+      canDownload={mode === 'action-creator'}
```
mlu-ssoft, line 40: move to `const canDownload = mode === 'action-creator'`, its repeated twice

libs/map/ui-results-view/src/lib/results-list/use-result.hook.ts
```diff
@@ -1,24 +1,23 @@
...
   const { comparisonModeEnabled, itemAddedToComparisonMode, canAddAsNewItemToComparisonMode, toggleCompareItem } =
     useComparisonMode();
 
-  const isSelected = useCallback((id: string) => selectedFeature?.id === id, [selectedFeature?.id]);
-
-  const isAddedToComparison = useCallback(
-    (item: TFeature) => itemAddedToComparisonMode(item),
-    [itemAddedToComparisonMode]
+  const isSelected = useCallback((id: string) => visibleFeature?.id === id, [visibleFeature]);
+  const isSelectedMultipleIndices = useCallback(
```
mlu-ssoft, line 14: add one empty line before for readibility reason

libs/map/ui-results-view/src/lib/results-list/result-item/multiple-items-action-buttons/multiple-items-action-buttons.component.tsx
```diff
@@ -0,0 +1,93 @@
...
+    comparisonEnabled,
+    itemAddedToComparisonMode,
+    canCompareItems,
+    downloadItem,
+    toggleCompareItem,
+    toggleItem,
+  } = useResult();
+  const [itemsInComparison, setItemsInComparison] = useState<string[]>([]);
+
+  const { thumbnail, ...indices } = feature.assets;
```
mlu-ssoft, line 29: use `useMemo` otherwise this will be new object every time and every reference that is using this will be rerendered

libs/map/ui-results-view/src/lib/results-list/result-item/multiple-items-action-buttons/multiple-items-action-buttons.component.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    <div>
+      <div className='flex space-between'>
+        {canDownload && <DownloadButton onDownload={() => downloadItem(feature)} disabled={comparisonEnabled} />}
+        <ToggleAssetsButton isOpened={isOpened} onToggle={onToggleShowAssets} itemsInComparison={itemsInComparison} />
+      </div>
+
+      {isOpened && (
+        <div className='mt-2 border-t border-t-bright-dark'>
+          {(Object.keys(indices) as TAssetName[]).map((key) => (
```
mlu-ssoft, line 76: maybe `Object.keys(feature.assets).filter(asset => asset !== 'thumbnail`)`? This: `const { thumbnail, ...indices } = feature.assets;` will be recreated on every change in this component and new instance for object will be created. I don't know if react will optimize it

libs/map/ui-results-view/src/lib/results-list/result-item/multiple-items-action-buttons/multiple-items-action-buttons.component.tsx
mlu-ssoft: or `const indicies = useMemo(() => Object.keys(feature.assets).filter(asset => asset !== 'thumbnail'), [feature.assets])`

libs/map/ui-results-view/src/lib/results-list/use-result.hook.ts
```diff
@@ -1,24 +1,23 @@
...
   const { mode } = useMode();
   const { comparisonModeEnabled, itemAddedToComparisonMode, canAddAsNewItemToComparisonMode, toggleCompareItem } =
     useComparisonMode();
 
-  const isSelected = useCallback((id: string) => selectedFeature?.id === id, [selectedFeature?.id]);
-
-  const isAddedToComparison = useCallback(
-    (item: TFeature) => itemAddedToComparisonMode(item),
-    [itemAddedToComparisonMode]
+  const isSelected = useCallback((id: string) => visibleFeature?.id === id, [visibleFeature]);
```
mlu-ssoft, line 13: whould this work:?

const isSelected = useCallback((id: string, assetName?: TAssetName) => {
if (assetName) {
return visibleFeature?.id === id && assetName === key assetNameWhichShouldBeDisplayed
}

return visibleFeature?.id === id;
}, [visibleFeature]);

if yes then there will be only one function for checking if something is selected

mlu-ssoft (changes requested):

mlu-ssoft:
Number of assets isn't shown after toggling on/off AC:
1. activate AC
2. go to history tab
3. load results
4. add two items to compare mode
5. enable compare mode
6. toggle off AC
7. toggle on AC

Badges for items aren't shown.

CleanShot 2025-02-11 at 17 27 16@2x

mlu-ssoft:
View button has wrong state after toggling off/on AC.

Steps:
1. activate AC
2. go to history tab
3. load results
4. click view on one of the assets
5. toggle off AC
6. toggle on AC

Right now button for selected asset has `View` text but it should be `Hide`.
Blue background should be visible.

CleanShot 2025-02-11 at 17 37 27@2x

mlu-ssoft:
Blank page (error) for ID: `5b37e23c-e3c4-11ef-ba1d`

CleanShot 2025-02-11 at 17 39 18@2x

libs/map/ui-results-view/src/lib/results-list/result-item/multiple-items-action-buttons/multiple-items-action-buttons.component.tsx
KatarzynaPaszko: `const indicies = useMemo(() => Object.keys(feature.assets).filter(asset => asset !== 'thumbnail'), [feature.assets])`
wont work, coz it will return an array of strings, and I need to return an object. 

I will memoize it

libs/map/ui-results-view/src/lib/results-list/result-item/multiple-items-action-buttons/multiple-items-action-buttons.component.tsx
KatarzynaPaszko: I have memoized `indices`

libs/map/ui-results-view/src/lib/results-list/result-item/multiple-items-action-buttons/multiple-items-action-buttons.component.tsx
```diff
@@ -0,0 +1,81 @@
...
+    (assetName: TAssetName) => {
+      return assetNameWhichShouldBeDisplayed === assetName && isSelected(feature.id);
+    },
+    [isSelected, feature.id, assetNameWhichShouldBeDisplayed]
+  );
+
+  useEffect(() => {
+    const count = countItemsAddedToComparisonMode(feature);
+    setItemsInComparison(count);
+  }, [itemAddedToComparisonMode, feature, countItemsAddedToComparisonMode]);
```
mlu-ssoft, line 46: I think `itemAddedToComparisonMode` in dependency array ins't needed. countItemsAddedToComparisonMode in dependency array has `comparisonItems.items` so it should retrigger every time new item will be added

libs/map/ui-results-view/src/lib/results-list/result-item/multiple-items-action-buttons/multiple-items-action-buttons.component.tsx
mlu-ssoft: move type (as TAssetName[]) there then

libs/map/ui-results-view/src/lib/results-list/results-list.component.tsx
```diff
@@ -69,16 +110,18 @@ export const ResultsList = ({ isFetching, features, hasNextPage, onLoadMore }: I
           dateTime={feature.properties.datetime}
           selected={isSelected(feature.id)}
           onToggleSelectedItem={() => toggleItem(feature)}
+          hasManyIndices={Object.keys(feature.assets).length > 1}
```
mlu-ssoft, line 113: it looks like the same condition as `Object.keys(feature.assets).length > 1` in ActionButtons? I would move it to separate function with domain name, eg: `cons hasManyIndices = (feature) => Object.keys(feature.assets).length > 1`. It has concrete meaning, in one case it you can read from the code what it is (here), in second case it is harder to guess (ActionButtons). Function can be placed outside both componens

mlu-ssoft (changes requested):

mlu-ssoft (approved):

PR #170, opened 2025-02-11
UKRIW-737 - Help in general view

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-769

- Extracted generating Help content to separate library
- Added Modal and  config for Help in general view


apps/eodh-fe/src/app/layout/help/translation-keys.tsx
```diff
@@ -0,0 +1,100 @@
+import { IHelpContent } from '@ukri/shared/ui/help';
+export const translationPath = 'MAP.GENERAL_HELP_MODAL';
+
+export const helpContentTranslationKeys: IHelpContent = {
```
mlu-ssoft, line 4: I would export `helpContentWithTranslations(helpContentTranslationKeys, translationPath)`. it will easier to understand what is happening and how content model is created. Furthermore, developer will have a choice - use this function or do it manually.

libs/shared/ui/help/src/lib/components/help-content.ts
```diff
@@ -0,0 +1,87 @@
+export const translationPath = 'MAP.ACTION_CREATOR_PANEL.HELP';
```
mlu-ssoft, line 1: this is path from AC, it shouldn't be here. This lib should be generic, it shouldn't have domain knowledge. It should be configured the same as generic help

mlu-ssoft (changes requested):

mlu-ssoft:
Table is not taking 100% width of the container

CleanShot 2025-02-11 at 17 00 14@2x 2

mlu-ssoft:
hash in url stays after page refresh

CleanShot 2025-02-11 at 17 00 14@2x

mlu-ssoft:
Help button in AC is not clickable. Should we remove it?

CleanShot 2025-02-11 at 16 57 41@2x

mlu-ssoft:
Last item shouldn't have border. This is common practice in lists.

CleanShot 2025-02-11 at 16 56 52@2x

mlu-ssoft:
Top padding (above title) is bigger than bottom one. It looks weird - I would assume that top/bottom padding should be the same.

CleanShot 2025-02-11 at 16 55 09@2x 2

mlu-ssoft:
Scroll should be positioned to the right (without right margin/padding)

CleanShot 2025-02-11 at 16 55 09@2x

KatarzynaPaszko:
> Help button in AC is not clickable. Should we remove it?

I didnt add it now. This button was added 5 months ago.

apps/eodh-fe/src/app/layout/help/help-button.component.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const HelpButton = ({ className }: { className?: string }) => {
+  const [helpIsOpen, setHelpIsOpen] = useState(false);
+
+  const openModal = useCallback(() => {
+    setHelpIsOpen(true);
+  }, []);
+
+  return (
+    <div className='z-40 flex'>
```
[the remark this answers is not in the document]
KatarzynaPaszko: Modal that is under this div is positioned absolute, and it needs z-index 40 in order to be displayed correctly

apps/eodh-fe/src/app/layout/help/translation-keys.tsx
KatarzynaPaszko: Perhaps  helpContentTranslationKeys is not best name as it will have also links urls and some image config, so it should be separated from translationPath. I will rename it to not be misleading

libs/map/feature-action-creator-panel/src/lib/content/help/help.component.tsx
```diff
@@ -1,42 +1,17 @@
...
-            content={memoizedIntroTranslation}
-            fontSize='medium'
-            fontWeight='regular'
-            className={helpStyles.helpIntro}
-          />
-          <TableOfContent />
-          <Answers />
-        </section>
+        <HelpContent
+          translationPath={translationPath}
```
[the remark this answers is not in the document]
KatarzynaPaszko: I have updated it so now Im passing just helpContentConfig that has inside path to translations as well as required object field

mlu-ssoft (approved):

PR #177, opened 2025-02-14
UKRIW-537 - No history message

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-537

### DIFF
--- a/libs/map/feature-action-creator-panel/src/lib/content/history/history.component.tsx
+++ b/libs/map/feature-action-creator-panel/src/lib/content/history/history.component.tsx
@@ -54,6 +54,17 @@ const LoadMoreButton = ({ isFetching, onClick }: ILoadMoreButtonProps) => {
   );
 };
 
+const NoHistory = () => {
+  return (
+    <div className='flex flex-col items-center p-4'>
+      <Error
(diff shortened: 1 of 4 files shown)


libs/map/feature-action-creator-panel/src/lib/content/history/history.component.tsx
```diff
@@ -87,9 +98,13 @@ export const History = () => {
     <Container>
       <Content>
         <section className='text-text-primary h-full overflow-x-visible overflow-y-scroll p-4'>
-          <div className='flex justify-end'>
-            <SortFilter onSortChange={changeOrder} sortKey={orderBy} />
-          </div>
+          {results.length === 0 ? (
```
mlu-ssoft, line 101: I would change it to:

`if (!results.length) {
return (<Container><Content><section className='text-text-primary h-full overflow-x-visible overflow-y-scroll p-4'><NoHistory /><section/></Content></Container>)
}`

you can also move section into separate component
1. mapping throug items doesn't make sense anyway - there is no items in the stock. So this condition is not only for filtering but it should change how and what is rendered in this component
2. moving condition to the top makes easier to understand - there are two different cases - with and without data

mlu-ssoft (changes requested):

mlu-ssoft (approved):

PR #182, opened 2025-02-17
UKRIW-806 - fixed thumbnail to be clickable on Search Mode

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-806

### DIFF
--- a/libs/map/ui-results-view/src/lib/results-list/result-item/result-item.component.tsx
+++ b/libs/map/ui-results-view/src/lib/results-list/result-item/result-item.component.tsx
@@ -38,12 +38,7 @@ export const ResultItem = ({
     [cloudCoverage]
   );
 
-  const disabled = useMemo(() => {
-    if (mode === 'action-creator' && hasManyIndices) {
-      return true;
-    }
-    return false;
-  }, [hasManyIndices, mode]);
+  const imageDisabled = useMemo(() => mode === 'action-creator' && hasManyIndices, [hasManyIndices, mode]);
 
   return (
     <div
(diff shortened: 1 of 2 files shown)


libs/map/ui-results-view/src/lib/results-list/result-item/result-item.component.tsx
```diff
@@ -36,14 +38,21 @@ export const ResultItem = ({
     [cloudCoverage]
   );
 
+  const disabled = useMemo(() => {
```
mlu-ssoft, line 41: you can simplify this: `const disabled = useMemo(() => mode === 'action-creator' && hasManyIndices, [hasManyIndices, mode]);`

mlu-ssoft: I would move this into `Image` component itself. This logic is related to image only, `disabled` name is to generic (in ResultItem component there may be many things disabled, eg ResultItem itself). Instead of passing disabled property I would pass `mode` and `hasManyIndices` then you could move this hook into `Image` component

mlu-ssoft (changes requested):

libs/map/ui-results-view/src/lib/results-list/result-item/result-item.component.tsx
KatarzynaPaszko: I have simplified it as suggested, but I have left it here, as `Image` component do not need to know why it is disabled, nor what is mode or how many indices result has.  In order to not confuse here what this disabled, I have renamed it to `imageDisabled` to make it very clear.

mlu-ssoft (approved):

PR #199, opened 2025-02-27
[UKRIW-822] feat: improve performance for Data Sets tree

mlu-ssoft:
Improve performance for Data Set tree. This partially solve issue - the issue still exists, but because of improved performance it is very edge case (user have to enable/disable AC 10-15 times)

Tickets:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-822


KatarzynaPaszko (commented):
Please update ticket in PR description from 746 to to 822

PR #209, opened 2025-03-05
[UKRIW-829] feat: add Planet Data Set

mlu-ssoft:
- add Planet Data Set
- add new Query Builder for Planet

Ticekts:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-829


KatarzynaPaszko (commented):
Comparison tool does not work for Planet data

KatarzynaPaszko (changes requested):
Compare Tool is not working for Planet Data

libs/map/data-access-stac-catalog/src/lib/query-builder/query.model.ts
```diff
@@ -34,6 +46,10 @@ export type TCatalogSearchParams = {
   dataSets: {
     public: {
       copernicus: TCopernicusSearchParams;
+      auxiliary?: object;
```
KatarzynaPaszko, line 49: Please add type for auxiliary object

libs/map/data-access-stac-catalog/src/lib/query-builder/workflow/workflow.query-builder.ts
KatarzynaPaszko, line 1: It seems that a lot of type definitions and methods are same in 
eodh-fe/libs/map/data-access-stac-catalog/src/lib/query-builder/comercial/planet.query-builder.ts
and
eodh-fe/libs/map/data-access-stac-catalog/src/lib/query-builder/workflow/workflow.query-builder.ts

Can you extract them to separate file and reuse

KatarzynaPaszko (approved):

libs/map/data-access-stac-catalog/src/lib/query-builder/workflow/workflow.query-builder.ts
mlu-ssoft: those are different QueryBuilders, each can have its own params/configuration
well probably we should have one `BaseQueryBuilder` class, then we don't need to repeat those types, but it will require more work. I will add todo and leave it as it is

PR #213, opened 2025-03-06
UKRIW-609 - AOI drawing tool pointer bug fix

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-609


libs/map/ui-map/src/lib/layers/aoi-layer/use-aoi-layer.hook.tsx
```diff
@@ -77,18 +85,46 @@ export const useAoiLayer = () => {
...
     return () => {
       map.removeInteraction(draw.draw);
     };
-  }, [map, draw, setShape, setDraw]);
+  }, [map, draw, setShape, setDraw, drawingTool, setDrawingTool]);
 
   useEffect(() => {
-    setDrawingTool(draw ? { enabled: true, type: draw.type } : undefined);
-  }, [draw, setDrawingTool]);
+    if (drawingTool?.type === 'rectangle') {
```
mlu-ssoft, line 98: use `switch-case` or `if-else`. Right now those ifs aren't connected (there is no if-else syntax)

mlu-ssoft (changes requested):

mlu-ssoft (approved):

PR #215, opened 2025-03-07
UKRIW-766 - time slider disabled state

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-766

### DIFF
--- a/libs/shared/ui/time-slider/src/lib/time-slider.styles.ts
+++ b/libs/shared/ui/time-slider/src/lib/time-slider.styles.ts
@@ -1,15 +1,15 @@
-export const sliderStyles = {
+export const sliderBgStyles = {
   container: 'h-[76px] w-full relative bg-background-main',
   innerContainer: 'w-full px-6 pb-4 pt-[17px] absolute top-0',
-  slider: `h-2
+};
+
+export const sliderStyles = (disabled?: boolean) =>  `h-2
   [&_.MuiSlider-track]:border-none
   [&_.MuiSlider-track]:h-[5px]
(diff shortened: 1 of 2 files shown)


libs/shared/ui/time-slider/src/lib/time-slider.styles.ts
```diff
@@ -21,5 +21,10 @@ export const sliderStyles = {
   [&_.MuiSlider-mark]:h-3
   [&_.MuiSlider-mark]:w-0.5
   [&_.MuiSlider-markLabel]:text-gray-500
-  [&_.MuiSlider-markLabel]:text-xs`,
-};
+  [&_.MuiSlider-markLabel]:text-xs 
+  ${
+    disabled
+      ? '[&_.MuiSlider-thumb]:bg-[#B4BBCA] [&_.MuiSlider-track]:bg-[#B4BBCA]'
+      : '[&_.MuiSlider-thumb]:bg-blue-600   [&_.MuiSlider-track]:bg-blue-600'
```
mlu-ssoft, line 28: can you use hex color instead of `bg-blue-600`? blue-600 is token from tailwind, not our design system. Noone will know its value and it will be hard to find it using plaintext search as it will be hard to figure it out that blue-number colors are used

mlu-ssoft (changes requested):

libs/shared/ui/time-slider/src/lib/time-slider.styles.ts
KatarzynaPaszko: updated

mlu-ssoft (approved):

PR #224, opened 2025-03-13
UKRIW-684 and UKRIW-565 - Notistack notification bug fixes

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-565
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-684


libs/shared/utils/notifications/src/lib/notifications.ts
```diff
@@ -0,0 +1,18 @@
+import { enqueueSnackbar } from 'notistack';
+
+const calculateDisplayTime = (message: string) => {
+  const wpm = 180; // readable words per minute
```
mlu-ssoft, line 4: variable/function names should be self-explainatory, additional comments shoulnd't be required. I would change names of variables to include information from comments

mlu-ssoft (changes requested):

libs/shared/utils/notification/project.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "notification",
```
mlu-ssoft, line 2: it should be `shared-utils-notification`. Do not change library files manually. The best is to remove library and create it once again. There is move command, but it sometimes doesn't updating files correctly

mlu-ssoft (changes requested):

mlu-ssoft (approved):

PR #240, opened 2025-03-19
UKRIW-575 - area node in ac bug fix

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-575

### DIFF
--- a/libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/value-node.component.tsx
+++ b/libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/value-node.component.tsx
@@ -38,7 +38,13 @@ export const formatUnit = (area: number, unit: TBaseUnit, t: TFunction, numberFo
   }
 };
 
-const formatArea = function (text: string, value: TAreaNode['value'], unit: TBaseUnit, t: TFunction, numberFormatting: string) {
+const formatArea = function (
+  text: string,
+  value: TAreaNode['value'],
+  unit: TBaseUnit,
+  t: TFunction,
+  numberFormatting: string
+) {
   const area = getArea(value);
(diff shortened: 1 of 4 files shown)


libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/value-node.component.tsx
```diff
@@ -28,12 +28,12 @@ export const formatUnit = (area: number, unit: TBaseUnit, t: TFunction) => {
   switch (value.unit.type) {
     case 'km2':
     case 'miles2': {
-      return `${value.value} ${t(value.unit.displayedValueTranslation)}<sup>2</sup>`;
+      return `${value.value.toLocaleString('en-US')} ${t(value.unit.displayedValueTranslation)}<sup>2</sup>`;
```
mlu-ssoft, line 31: locale should be configurable, I would move this into `shared/utils/settings` `useSettings` hook

libs/map/feature-action-creator-panel/src/lib/content/workflow/nodes/area/value-node.component.tsx
```diff
@@ -28,12 +28,12 @@ export const formatUnit = (area: number, unit: TBaseUnit, t: TFunction) => {
...
     }
 
     case 'km':
     case 'miles': {
-      return `${value.value} ${t(value.unit.displayedValueTranslation)}`;
+      return `${value.value.toLocaleString('en-US')} ${t(value.unit.displayedValueTranslation)}`;
```
mlu-ssoft, line 36: locale should be configurable, I would move this into `shared/utils/settings` `useSettings` hook

mlu-ssoft (changes requested):

mlu-ssoft (approved):

PR #243, opened 2025-03-20
[UKRIW-868] feat: add loading spinner for thumbnails

mlu-ssoft:
This is part of UKRIW-868 ticket. It is adding only spinner for thumbnails.

Tickets:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-868


libs/map/ui-results-view/src/lib/results-list/result-item/image.component.tsx
KatarzynaPaszko, line 1: It doesn't work on Search View

KatarzynaPaszko (changes requested):

PR #246, opened 2025-03-20
[UKRIW-868] feat: add spinner for thumbnails

mlu-ssoft:
This is part of UKRIW-868 ticket. It is adding only spinner for thumbnails.

Tickets:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-868


KatarzynaPaszko (commented):
It doesnt work as it should You can still see no img icon instead of spinner

KatarzynaPaszko:
> It doesnt work as it should You can still see no img icon instead of spinner

ok, I retried and it works

libs/map/ui-results-view/src/lib/results-list/result-item/image.component.tsx
KatarzynaPaszko, line 1: I think it is worth to extract w-[132px] h-[132px] min-w-[132px] min-h-[132px] to const as this styling is used in many places

libs/map/ui-results-view/src/lib/results-list/result-item/image.component.tsx
```diff
@@ -11,14 +20,15 @@ interface IImageProps {
 
 export const Image = ({ imageUrl, onToggle, disabled = false }: IImageProps) => {
   const { mode } = useMode();
-  const [displayError, setDislayError] = useState(false);
   const [imageSrc, setImageSrc] = useState<string | undefined>(undefined);
+  const [status, setStatus] = useState<TImageStatus>('loading');
+  const isLoading = useMemo(() => status === 'loading' || !imageSrc, [status, imageSrc]);
```
KatarzynaPaszko, line 25: If for some reason `imageSrc` wont be returned from BE, then it will keep on showing loading spinner, which is not correct. In that case we should display image for status error

mlu-ssoft: but it works like that, there is catch in useEffect which set status with error. If url will be null the it won't be fetched and error will be thrown

KatarzynaPaszko (approved):

PR #251, opened 2025-03-21
UKRIW-754 - Delete workflow functionality

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-754


libs/map/feature-action-creator-panel/src/lib/content/history/history-tile/history-tile.component.tsx
```diff
@@ -39,15 +28,27 @@ export const HistoryTile = ({
...
   className,
-  selected,
-  children,
-}: PropsWithChildren<IHistoryTileProps>) => {
+  selectedResult,
+  loadResultsStatus,
+  onHide,
+  onShow,
+}: IHistoryTileProps) => {
+  const [deleteInProgress, setDeleteInProgress] = useState(false);
```
mlu-ssoft, line 36: a lot is happening in this component, I would move logic to hook, right now this component is mixed with logic and presentation

libs/map/feature-action-creator-panel/src/lib/content/history/history-tile/delete-item.component.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  useEffect(() => {
+    if (isError) {
+      displayNotification(t(`${translationPath}.ERROR`), 'error');
+    }
+  }, [isError, t]);
+
+  return (
+    <div>
+      <Notification type='errorLight' className='my-4 p-4 shadow-none' closeButtonVisible={false}>
```
mlu-ssoft, line 75: I would change type to custom. `errorLight` is used only here, we don't know if we will have more places where this notification type will be displayed. It looks as something domain specific which will be used in only one place. I would change Notification type to allow `custom` property in which then there will be no color used - color should be passed as `className`. This will allow us to add more variations which are domain specific

libs/map/feature-action-creator-panel/src/lib/content/history/history-tile/delete-item.component.tsx
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  if (isPending) {
+    return (
+      <>
+        <Button size='medium' appearance='text' text={`${translationPath}.NO_CTA`} onClick={onNoClick} disabled />
+        <Button
+          size='medium'
+          text={
+            <>
```
mlu-ssoft, line 29: wouldn't it be better to add it as a `span` tag? Right now you have to know how `text` is handled inside button, eg is it displayed using some positioning. If implementation will change, this may broke

mlu-ssoft (changes requested):

mlu-ssoft (approved):

PR #258, opened 2025-03-21
UKRIW-876 - Clipping notification

KatarzynaPaszko:
https://ukri-spyrosoft.atlassian.net/browse/UKRIW-876


libs/map/feature-action-creator-panel/src/lib/content/workflow/workflow.component.tsx
```diff
@@ -90,16 +93,23 @@ export const Workflow = () => {
...
+          functions: functionNodes,
+        },
+        functions: data,
       },
-      functions: data,
-    });
-  }, [data, getNodesByType, mutate]);
+      {
+        onSuccess: () => {
+          displayNotification(t('MAP.ACTION_CREATOR_PANEL.WORKFLOW.INFO.CLIPPING_NOTIFICATION'), 'default');
```
mlu-ssoft, line 108: you should detect if clipping function isn't added. This message should be shown only if in functions clipping function wasn't chose

mlu-ssoft (changes requested):

mlu-ssoft (approved):

PR #259, opened 2025-03-24
[UKRIW-881] feat: use default workspace

mlu-ssoft:
Tickets:
- https://ukri-spyrosoft.atlassian.net/browse/UKRIW-881


mlu-ssoft:
this is not how it works, I close this PR as the mechanism for getting default workspace is different

